Issue #132, opened 2015-04-05
Implement missing Char Methods


geoffreymcgill:
https://msdn.microsoft.com/en-us/library/System.Char_methods%28v=vs.110%29.aspx

	IsControl
	IsDigit
	IsHighSurrogate
	IsLetter
	IsLetterOrDigit
	IsLower
	IsLowSurrogate
	IsNumber
	IsPunctuation
	IsSeparator
	IsSurrogate
	IsSurrogatePair
	IsSymbol
	IsUpper
	IsWhiteSpace

PR #158, opened 2015-04-08
Added Char.IsDigit

Dim0N22:
Hello. I'm new to Bridge but I think it is a very interesting framework.

I've tried to implement Char.IsDigit function, but I really don't see how do you get minimized bridge.js file. 

Could you help, please? I'll continue to implement Char functions. 

Thank you.

### DIFF
--- a/Bridge/Bridge.csproj
+++ b/Bridge/Bridge.csproj
@@ -159,6 +159,7 @@
   </ItemGroup>
   <ItemGroup>
     <Content Include="Resources\Array.js" />
+    <Content Include="Resources\Char.js" />
     <Content Include="Resources\Finally.js" />
     <Content Include="Resources\Init.js" />
     <Content Include="Resources\Attribute.js" />
--- a/Bridge/Resources/Bridge.jsb
+++ b/Bridge/Resources/Bridge.jsb
@@ -4,6 +4,7 @@
   <file name="Core.js" />
   <file name="Nullable.js" />
   <file name="String.js" />
+  <file name="Char.js" />
   <file name="Class.js" />
   <file name="Exception.js" />
   <file name="Interfaces.js" />
@@ -29,6 +30,7 @@
     <include name="Core.js" />
     <include name="Nullable.js" />
     <include name="String.js" />
+    <include name="Char.js" />
     <include name="Class.js" />
     <include name="Exception.js" />
     <include name="Interfaces.js" />
--- a/Bridge/Resources/Char.js
+++ b/Bridge/Resources/Char.js
@@ -1,0 +1,15 @@
+﻿// @source Char.js
+
+(function () {
+    var chr = {
+        isDigit: function (str) {
+            if (str.length > 1) {
+                return false;
+            }
+
+            return str >= "0" && str <= "9";
+        }
+    };
+
+    Bridge.Char = chr;
+})();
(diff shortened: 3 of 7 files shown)


Dim0N22:
Sorry, I've found Knowledge Base on bridge.net. Still it would be nice to have a link from GitHub.

geoffreymcgill:
Thanks for `Char.IsDigit`. We'll review right away. To build `bridge.js` and `bridge.min.js`, double-click the following file (in your local File Explorer). 

https://github.com/bridgedotnet/Bridge/blob/master/Bridge/Resources/build.bat

Dim0N22:
Found it just minutes before you created the comment.

Thank you for help!

geoffreymcgill:
@vladsch we should just add this `build.bat` call to the Pre-build step. I've been tripped up a couple times by forgetting to run this file. It's so quick to run that adding to Pre-build should not be noticeable.

Dim0N22:
I think it's better to create separate Char object in js. What do you think?

Mimisss:
@Dim0N22 It will be needed sooner or later especially to implement the rest of the Char class methods like IsControl, IsLetter, IsNumber, etc.

geoffreymcgill:
Yes, a separate **Char.js** file is required, and we can get working on a `Bridge.Char.*` API.

geoffreymcgill:
@Mimisss There is some overlap in the work being done in this thread with Issue #132. Can you coordinate?

Dim0N22:
@Mimisss just tell what part should I focus on.

Mimisss:
@geoffreymcgill No, fortunately there is not. 

@Dim0N22 Feel free to work on the Char methods mentioned in #132.

Mimisss:
@Dim0N22 

Tip 1: `char` is treated as its numeric value. There is `String.fromCharCode` to get the corresponding character.

Tip 2: The `[Template("({ch} >= 48 && {ch} <= 57)")]` "implementation" of the `Char.IsDigit` method is a nice "shortcut" way to implement methods. 

Tips 1 & 2 example: This: 
`Console.WriteLine(char.IsDigit('5'));`

is translated to:

`console.log((53 >= 48 && 53 <= 57));`

Dim0N22:
@Mimisss
Tip 1: Thank you, I will investigate it.

Tip 2: I think it is more consistent to have all the ```Char``` logic in ```Char``` object. Having logic in the ```Template``` is a bit frustrating and not what an user can expect. Anyway you decide.

Mimisss:
@Dim0N22 Tip 2: Absolutely. I mentioned `Template` because it can be handy in some situations; with overloaded methods, for example, where all you want is to call the same function defined in `Char.js` with adjusted parameters.

geoffreymcgill:
Hi @Dim0N22, 

Just to let you know we haven't forgotten about this pull request. I will be reviewing the current status of the Char class tomorrow (Thursday). 

Hope all is well with you. Feel free to contact me (geoff@object) directly any time if you have questions or comments. 

Thanks again for contributing this pr.

Dim0N22:
@geoffreymcgill Hi.

Thank you for your time! 
I am afraid I have no idea how to implement Char in JS with handling all the unicode characters. Probably you have an idea?

Mimisss:
This is the best resource I've found so far: 
http://xregexp.com/plugins/

Such a solution is needed to categorize characters beyond ASCII.

Mimisss:
Char methods have been added to the master repo.

PR #268, opened 2015-06-12
Building on Linux

wind39:
I tried to build Bridge on both inside MonoDevelop and using xbuild on console. Both can't build with the following error:

```
Building: Bridge (Debug)

Build started 11/06/2015 22:09:11.
__________________________________________________
Project "/work/projects/Repositories/Bridge/Bridge/Bridge.csproj" (Build target(s)):
	
	Target PrepareForBuild:
		Configuration: Debug Platform: AnyCPU
	
	Target PreBuildEvent:
		Executing: "/work/projects/Repositories/Bridge/Bridge//Resources/build.bat"
		/work/projects/Repositories/Bridge/Bridge//Resources/build.bat: 1: /work/projects/Repositories/Bridge/Bridge//Resources/build.bat: %~dp0\JSBuilder\JSBuildConsole.exe: not found
		/work/projects/Repositories/Bridge/Bridge//Resources/build.bat: 3: /work/projects/Repositories/Bridge/Bridge//Resources/build.bat: move: not found
		/work/projects/Repositories/Bridge/Bridge//Resources/build.bat: 4: /work/projects/Repositories/Bridge/Bridge//Resources/build.bat: move: not found
/usr/lib/mono/4.5/Microsoft.Common.targets: error : Command '"/work/projects/Repositories/Bridge/Bridge//Resources/build.bat"' exited with code: 127.
	Task "Exec" execution -- FAILED
	Done building target "PreBuildEvent" in project "/work/projects/Repositories/Bridge/Bridge/Bridge.csproj".-- FAILED
	
Done building project "/work/projects/Repositories/Bridge/Bridge/Bridge.csproj".-- FAILED

Build FAILED.
Errors:

/work/projects/Repositories/Bridge/Bridge/Bridge.csproj (Build) ->
/usr/lib/mono/4.5/Microsoft.Common.targets (PreBuildEvent target) ->


---------------------- Done ----------------------

Build: 1 error, 0 warnings
```

With the proposed solution files, I'm able to build Bridge on Linux.

PR #270, opened 2015-06-15
Building on Linux without multiple .csproj files

wind39:
@fabriciomurta Following your instructions in [#268](https://github.com/bridgedotnet/Bridge/pull/268), I developed a better approach.
Please ignore [#268](https://github.com/bridgedotnet/Bridge/pull/268).


fabriciomurta:
Are you able to run this on linux?
`../../Resources/JSBuilder/JSBuildConsole.exe /path="../../Resources/Bridge.jsb"`

I was expecting (at least) a `mono (...)JSBuildConsole.exe` or an a.out version of this. :)

fabriciomurta:
I expect you to get a
`bash: ./Resources/JSBuilder/JSBuildConsole.exe: cannot execute binary file`

Every time you try to run this win32 file on linux or OS X. That would not be a problem if you're on mingw or cygwin though.

fabriciomurta:
Alright, I can run it with `mono`.

PR #287, opened 2015-07-08
Fix string.Concat for multiple arguments

txdv:
If the string.Concat with params[] string is taken
the javascript implementation will take only the
first argument. The function basically returns the
first argument, since everything else is ommited.

I described it a bit more detailed in the comments of the commit https://github.com/bridgedotnet/Bridge/commit/43ba944e8e44607057fb351125b5817bc85d58a8

### DIFF
--- a/Bridge/Resources/String.js
+++ b/Bridge/Resources/String.js
@@ -307,15 +307,15 @@
             return arr;
         },
 
-        concatArray: function (strs) {
-            if (strs == null) {
+        concatArray: function () {
+            if (arguments.length == 0) {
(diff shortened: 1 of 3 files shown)


txdv:
I'm closing the pull request because this is a tradeoff anyway and there is no easy fix. Bridge.NET 2.0 will probably fix this in the manner that Saltarelle does.

fabriciomurta:
Well, but I believe that at least we should open an issue to track and (hopefully) fix this.

fabriciomurta:
Hello, I believe @Mimisss went ahead and fixed this on #295. :)

PR #371, opened 2015-08-19
Fix wrong xml param name.

txdv:
This won't compile with the mono compiler, because the
whitespace in front doesn't get omitted with the mono
compiler.

Also, the recompiled version of bridge.js and bridge.min.js is quite different: https://gist.github.com/txdv/9666bf169561a866846c on mono, but this has no code change, only comment change, so the generated output on windows should be the same.


DaniilVeriga:
Interesting.

Thank you very much for the contribution!

Issue #264, opened 2015-06-01
Global.Location.Hash is read only. Should it be?


Mimisss:
Or should it be read/write?

geoffreymcgill:
What does the Mozilla spec say?

Mimisss:
> Though Window.location is a read-only Location object, you can also assign a DOMString to it. 

https://developer.mozilla.org/en-US/docs/Web/API/Window/location

fabriciomurta:
As I read, there's kind of a syntax sugar in that, so anything changing the location (would be the same as changing location.href), should force a page reload to the new location so... I guess it really should be made an exception and allow assigning and being used like a string... The side effect would be treated browser-side.

DaniilVeriga:
Well, according to the Mozilla docs all the Location's fields are not read only except `origin`.
https://developer.mozilla.org/en-US/docs/Web/API/Location

In Bridge.NET we currently have all fields read only except `href`. I think it is a defect and going to fix it.

geoffreymcgill:
@DaniilVeriga  - It appears this issue was solved, and should have been closed. Can you investigate?

DaniilVeriga:
@geoffreymcgill, you are right. The fix was included to 1.9.

PR #402, opened 2015-08-28
[FIX] Location.Hash and other fields are read only, but they shouldn't be (except Origin).

DaniilVeriga:
Relates to #264.

I am relying on the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/Location).


LeonidVeriga:
Test will be done after merge as it cause Testing repo not compiling

DaniilVeriga:
The test has been added in [this commit](https://github.com/bridgedotnet/Testing/commit/aa3267b9d2a690ac46215e7481cc9a2f14527682).

Issue #406, opened 2015-09-01
[SWITCH] More than 7 string case statements causes compilation errors


LeonidVeriga:
http://forums.bridge.net/forum/bridge-net-pro/bugs/597-switch-statement-causes-c-compiler-to-crash

>This seems to only be a problem when the switch statement is matching against strings and when there are more than a certain number of case statements, 7 in my case.

The issue is not reproducible in VS2015.

```C#
using Bridge.Html5;

namespace BridgeTest
{

    public class App
    {
        public const string TESTA = "TESTA";
        public const string TESTB = "TESTB";
        public const string TESTC = "TESTC";
        public const string TESTD = "TESTD";
        public const string TESTE = "TESTE";
        public const string TESTF = "TESTF";
        public const string TESTG = "TESTG";

        [Ready]
        public static void Main()
        {
            var test = TESTD;
            switch (test)
            {
                case TESTA:
                    Global.Alert("EnumA");
                    break;

                case TESTB:
                    Global.Alert("EnumB");
                    break;

                case TESTC:
                    Global.Alert("EnumC");
                    break;

                case TESTD:
                    Global.Alert("EnumD");
                    break;

                case TESTE:
                    Global.Alert("EnumE");
                    break;

                case TESTF:
                    Global.Alert("EnumF");
                    break;

                case TESTG:
                    Global.Alert("EnumG");
                    break;
            }
        }
    }
}
```

erik-kallen:
It's likely because some member of ```System.Collections.Generic.Dictionary<TKey, TValue>``` member is missing.

PR #417, opened 2015-09-03
[FIX] More than 7 string case statements causes compilation errors #406

vladsch:
https://github.com/bridgedotnet/Bridge/issues/406


LeonidVeriga:
Vlad, the PR missed a test - I added it. `bridge.js` and `bridge.min.js` should have been updated as well as new code added.

There is a couple things I'm concerned about:
1) index.html tests fail (from Bridge.Test sln)
    The following calls:
```C#
var d1 = new Dictionary<int, string>(10);

var c = new TestEqualityComparer();
var d2 = new Dictionary<string, string>(10, c);
```
   get transpiled:
```JavaScript
var d1 = new Bridge.Dictionary$2(Bridge.Int,TValue)();

var c = ...
var d2 = new Bridge.Dictionary$2(String, String)(null, c);
```

2) there is a new function `getCurrent$1`added into `bridge.js` that looks like the same as `getCurrent`
```JavaScript
getCurrent: function () {
    return this.$getCurrent.call(this.scope);
},

getCurrent$1: function () {
    return this.$getCurrent.call(this.scope);
},
```

Issue #418, opened 2015-09-04
Emit/output error when using += operator with Delegates


fabriciomurta:
A severe output error happens whenever an Delegate is bound using the += operator.

The bug was reported here [Delegate Compilation Bug](http://forums.bridge.net/forum/bridge-net-pro/bugs/614)

For instance:
- broken: `test.Delegate += Callback;`
- works: `test.Delegate = test.Delegate + Callback;`

fabriciomurta:
This issue might be related to #407.

PR #424, opened 2015-09-06
[FIX][#418] Emit/output error when using += operator with Delegates

vladsch:
https://github.com/bridgedotnet/Bridge/issues/418


LeonidVeriga:
The following code does not get transpiled with error `startIndex cannot be larger than length of string`:
```C#
using Bridge;

namespace Demo
{
    public class Bridge418
    {
        public delegate int MyDelegate(int data);

        public MyDelegate Delegate { get; set; }

        public int CallDelegate(int data)
        {
            return Delegate(data);
        }
    }

    public class App
    {
        [Ready]
        public static void Main()
        {
            var t = new Bridge418();
            t.Delegate += (i) => i * 2;
        }
    }
}
```

Issue #467, opened 2015-09-16
Call to base.GetHashCode() causes compilation to fail


vladsch:
http://forums.bridge.net/forum/bridge-net-pro/bugs/678-call-to-base-gethashcode-causes-compilation-to-fail

PR #468, opened 2015-09-16
[FIX][#467] Call to base.GetHashCode() causes compilation to fail

vladsch:
https://github.com/bridgedotnet/Bridge/issues/467


LeonidVeriga:
Vlad, it goes in endless loop due to in bridge.js `getHashCode`
```if (Bridge.isFunction(value.getHashCode)) {
                return value.getHashCode();
            }
```

Please see test `ClientTestLibrary.TestBridgeIssues.N467()` (rebuild Bridge.Frameworks.Builder.Testing and run index.html)

Issue #477, opened 2015-09-18
Methods' optional parameters are not taken into account in generated JavaScript


DaniilVeriga:
http://forums.bridge.net/forum/bridge-net-pro/bugs/680

It might be controversial a bit it is a defect or a feature, but in my opinion it is rather a defect.

A similar TypeScript-related defect - #292.

PR #480, opened 2015-09-19
[FIX][#477] Methods' optional parameters are not taken into account in generated JavaScript

vladsch:
https://github.com/bridgedotnet/Bridge/issues/477


LeonidVeriga:
Vlad, the following code `var s = string.Format("x");` translates to `var s = Bridge.String.format("x", );` please note `"x", `

Issue #488, opened 2015-09-22
CRLF line endings being added to bridge.js generated by the build.bat script in some git clients


fabriciomurta:
With the fix in #387, the new file merging script had a setting converted during checkout of the repo in some git clients, adding \n to the header.

Turn the '.txt' file (`unixnl.txt`) into a binary file (`unixnl.dat`) to avoid git to keep translating it to/from LF<=>CRLF  and use only LF as line ending (the only character the file has).

Also transform the other *.txt files under dostools so they are treated as binary files by Git.

PR #489, opened 2015-09-22
Renames build.bat's accessory files so git treat them as binary. [#488]

fabriciomurta:
This will hopefully eliminates the mixed line endings in Bridge.JS for certain git clients!


LeonidVeriga:
@fabriciomurta, if I get a fresh clone then *unixnl.dat* contains CRLF (TortoiseGIT)

fabriciomurta:
Thanks for the feedback @LeonidVeriga ! I was thinking about that already. Might need `*.dat binary` in `.gitattributes`. Let's give it a shot.

PR #504, opened 2015-10-05
Add Math.Sign(double) prototype

gogglespisano:


LeonidVeriga:
Hi @gogglespisano. Thanks for your Pull Request. I created a separate one #515 based on your changes but for branch release1.10 as it is a source for the next 1.10 release. I had to change the code to support sign() method for IE by this change, a QUnit test added as well.

PR #530, opened 2015-10-22
Add AngularJS project to Frameworks references. [bridgedotnet/Framewo…

clrarchitect:
…rks#1]


fabriciomurta:
Sorry. I've given you a wrong direction in the other comment. You can switch to the `AngularJS` branch on the `Bridge` project as well. There, the reference is already made. As AngularJS is still on a specific branch, we can't add the reference in the Bridge repository.

Issue #548, opened 2015-11-02
TypeArrays miss the [Name("TypedArray")] attribute


DaniilVeriga:
The issue is discovered due to this [comment](https://github.com/bridgedotnet/Bridge/issues/182#issuecomment-152838336).

**C#**: `var float32Array = new Float32Array();`

**JavaScript**:

 * Actual: `var float32Array = new Bridge.Html5.Float32Array();` - it throws *Uncaught TypeError: Cannot read property 'Float32Array' of undefined*.
 * Expected: `var float32Array = new Float32Array();`

The [live example](http://live.bridge.net/#5baecb22616d86c1ba52).

It is actual for all the TypedArray classes.

fabriciomurta:
There's no Float32Array() constructor taking 0 arguments.

The valid overloads are:
```csharp
        public Float32Array(Any<int, long, uint, ulong> length);
        public Float32Array(Float32Array typedArray);
        public Float32Array(object objectInstance);
        public Float32Array(ArrayBuffer buffer, uint byteOffset = 0, uint length = 0);
```

Sources:
[MDN reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float32Array)
[Bridge source for Float32Array](https://github.com/bridgedotnet/Bridge/blob/master/Html5/TypedArray/Float32Array.cs)

DaniilVeriga:
Interesting why it compiles in Live Bridge. Okay, it is probably an issue with Live Bridge. Another story.

Another story more is that `new Float32Array()` actually works in browsers - tested in Chrome, Firefox and IE. But yes, it is not mentioned in the docs and I don't think we should add it.

Please forget the Float32Array() constructor taking 0 arguments. My bug report wasn't too accurate.

Now lets talk about all the available constructors. Do they work?:) As far as I can see they don't. The reason is probably missing a `[Name("Float32Array")]` attribute on the Float32Array class.

fabriciomurta:
The information above was just a side note, it did not invalidate the surplus `Bridge.Html5.` namespace reference on class instantiation.

Issue #549, opened 2015-11-02
TypedArray.Prototype should not be in C# and in emitted JavaScript


DaniilVeriga:
The issue is discovered due to this [comment](https://github.com/bridgedotnet/Bridge/issues/182#issuecomment-152838336).

**C#**: 
* Actual: `float32Array.Prototype.Buffer`
* Expected: `float32Array.Buffer`

The `.Prototype` thing should not be required in C# code.

**JavaScript**:
* Actual: `float32Array.prototype.buffer` - it throws an error that `prototype` is not defined.
* Expected: `float32Array.buffer`

The [live example](http://live.bridge.net/#0fce2de66ff629f504f6).

It is actual for all the TypedArray classes.

LeonidVeriga:
[Test results on Win10 + Edge](https://saucelabs.com/beta/tests/994064168b2749e797668e3a89f2fec1/jsunit)
```
#549 - TestUseCase
Float32Array ToLocaleString
Expected:	0,11,0,0,0,5,0,0,0,99
Result:	0.00, 11.00, 0.00, 0.00, 0.00, 5.00, 0.00, 0.00, 0.00, 99.00

Float64Array ToLocaleString
Expected:	0,11,0,0,0,5,0,0,0,99
Result:	0.00, 11.00, 0.00, 0.00, 0.00, 5.00, 0.00, 0.00, 0.00, 99.00

Int16Array ToLocaleString
Expected:	0,11,0,0,0,5,0,0,0,99
Result:	0.00, 11.00, 0.00, 0.00, 0.00, 5.00, 0.00, 0.00, 0.00, 99.00

Int32Array ToLocaleString
Expected:	0,11,0,0,0,5,0,0,0,99
Result:	0.00, 11.00, 0.00, 0.00, 0.00, 5.00, 0.00, 0.00, 0.00, 99.00

Int8Array ToLocaleString
Expected:	0,11,0,0,0,5,0,0,0,99
Result:	0.00, 11.00, 0.00, 0.00, 0.00, 5.00, 0.00, 0.00, 0.00, 99.00

Uint16Array ToLocaleString
Expected:	0,11,0,0,0,5,0,0,0,99
Result:	0.00, 11.00, 0.00, 0.00, 0.00, 5.00, 0.00, 0.00, 0.00, 99.00

Uint32Array ToLocaleString
Expected:	0,11,0,0,0,5,0,0,0,99
[...]

PR #560, opened 2015-11-03
Removes 'Bridge.Html5.' namespace ref from TypedArray classes

fabriciomurta:
This implements the fix for issue #548


fabriciomurta:
This should be merged with issue #549. Else, headaches due to conflicts during merge.

PR #561, opened 2015-11-03
Fixes TypedArrays so they inherit not append its prototype methods

fabriciomurta:
This fixes issue #549.


fabriciomurta:
This should be merged to issue #549, or else... merging headaches.

Issue #555, opened 2015-11-03
Review String.Substring implementation


geoffreymcgill:
Functionality may differ from the C# implementation:

Related forum discussion:
http://forums.bridge.net/forum/general/open-discussions/879

Issue #554, opened 2015-11-03
Implement String.Remove


geoffreymcgill:
Implement String.Remove and any overloads

https://msdn.microsoft.com/en-us/library/9ad138yc%28v=vs.110%29.aspx

Related forum discussion:
http://forums.bridge.net/forum/general/open-discussions/879

geoffreymcgill:
Related to #385

Issue #385, opened 2015-08-21
Review String implementation


geoffreymcgill:
We keep running into missing (#381) String methods. A full review and comparison is required.

https://msdn.microsoft.com/en-us/library/System.String_methods%28v=vs.110%29.aspx

Mostly just a comparison of C# Methods to the Bridge implementation to ensure we have coverage. Add any missing Methods, and confirm all overloads are available.

geoffreymcgill:
Another String implementation issue: #554

geoffreymcgill:
Issue with String.Substring: #555

LeonidVeriga:
List of added methods:
```
public String(char[] value, int startIndex, int length);
 
public extern int LastIndexOf(char ch);
public extern int LastIndexOf(string subString);
public extern int LastIndexOf(string subString, int startIndex);
public extern int LastIndexOf(char ch, int startIndex, int count);
public extern int LastIndexOf(string subString, int startIndex, int count);
public extern int LastIndexOf(char ch, int startIndex);
public extern int LastIndexOfAny(params char[] ch);
public extern int LastIndexOfAny(char[] ch, int startIndex);
public extern int LastIndexOfAny(char[] ch, int startIndex, int count);
public extern string Replace(char oldChar, char replaceChar);
public extern string[] Split(char[] separator);
public extern string[] Split(char[] separator, int limit);
public extern string[] Split(char[] separator, int limit, StringSplitOptions options);
public extern string[] Split(char[] separator, StringSplitOptions options);
public extern string[] Split(string[] separator, StringSplitOptions options);
public extern string[] Split(string[] separator, int limit, StringSplitOptions options);
public extern string[] Split(string separator);
public extern string[] Split(char separator);
public extern string[] Split(Regex regex);
public extern string[] Split(char separator, int limit);
public extern string[] Split(Regex regex, int limit);
public extern string[] Split(string separator, int limit);
public string JsSubstring(int indexA)
public string JsSubstring(int indexA, int indexB)
public string Substring(int start)
public string Substring(int start, int length)
public extern string Trim(params char[] values);
public extern string TrimStart(params char[] values);
public extern string TrimEnd(params char[] values);
public extern string TrimStart();
public extern string TrimEnd();
public extern int CompareTo(string other);
public extern string Insert(int startIndex, string value);
public static extern string Join(string separator, params string[] args);
public static extern string Join(string separator, params object[] args);
public static extern string Join(string separator, IEnumerable<string> args);
[...]

Issue #381, opened 2015-08-21
string.Join does not work as expected


DanTup:
In my Bridge project in VS the `string` class appears to have no `Join` method. This code fails to compile:

```
var s = string.Join("", new[] { "a", "b" });
```

> 'string' does not contain a definition for 'Join'

Strangely, if I [try it on live.bridge.net](http://live.bridge.net/#029f86e0e7470ff1bf5b) the code compiles, but the results are not as expected. This code:

```
var s = string.Join("", new[] { "a", "b" });
Global.Alert(s);
```

produces this JS:

```
var s = Bridge.Linq.Enumerable.from(String).join("", ["a", "b"], null, null);
Bridge.global.alert(s);
```

which alerts `[object Object]` instead of `ab`.

geoffreymcgill:
A defect or a feature? ;)

geoffreymcgill:
Five overloads are required:

`(String)` - https://msdn.microsoft.com/en-us/library/dd783876(v=vs.110).aspx
`(String, IEnumerable<T>)` - https://msdn.microsoft.com/en-us/library/dd992421(v=vs.110).aspx
`(String, Object[])` - https://msdn.microsoft.com/en-us/library/dd988350(v=vs.110).aspx
`(String, String[])` - https://msdn.microsoft.com/en-us/library/57a79xd0(v=vs.110).aspx
`(String, String[], Int32, Int32)` - https://msdn.microsoft.com/en-us/library/tk0xe5h0(v=vs.110).aspx

LeonidVeriga:
Hi @DanTup, we have reviewed string implementation and fixed this issue. Thanks for reporting!

PR #564, opened 2015-11-05
Review String implementation

vladsch:
https://github.com/bridgedotnet/Bridge/issues/381
https://github.com/bridgedotnet/Bridge/issues/385
https://github.com/bridgedotnet/Bridge/issues/554
https://github.com/bridgedotnet/Bridge/issues/555


Bridge/Resources/String.js
```diff
@@ -320,6 +353,53 @@
             var reg = new RegExp(Bridge.String.escape(a), "g");
 
             return str.replace(reg, b);
+        },
+
+        insert: function (idx, strA, strB) {
```
Suchiman: This method will fail with "index is undefined": parameter is named idx but method uses index

Issue #582, opened 2015-11-10
Missed DateTime API


DaniilVeriga:
http://forums.bridge.net/forum/general/feature-requests/724

`public DateTime ToUniversalTime()` - FYI: it is already in v2.
`public DateTime Add(TimeSpan value)`
`public DateTime AddTicks(long value)`
`public DateTime Subtract(TimeSpan value)`
`public TimeSpan TimeOfDay { get; }`
`public long Ticks { get; }`

A related Bridge v2 [issue](https://github.com/bridgedotnet/Bridge2/issues/36).

LeonidVeriga:
breaking-change: `DateTime.ToUTC` renamed to `DateTime.ToUniversalTime`

PR #612, opened 2015-11-15
Issue582 Missed DateTime API

vladsch:
https://github.com/bridgedotnet/Bridge/issues/582


LeonidVeriga:
The following methods/properties implemented:
```
public DateTime ToUniversalTime()
public DateTime ToLocalTime()
public extern DateTime Add(TimeSpan value);
public extern DateTime AddTicks(long value);
public extern DateTime Subtract(TimeSpan value);
public TimeSpan TimeOfDay { get; }
public long Ticks {get; }
```

Issue #645, opened 2015-11-24
Bridge 1.10.0 subsequent builds might fail


DaniilVeriga:
http://forums.bridge.net/forum/bridge-net-pro/bugs/1068

PR #646, opened 2015-11-24
[FIX] [#645] Bridge 1.10.0 subsequent builds might fail

DaniilVeriga:
#645


LeonidVeriga:
Issue:
Microsoft.Build.Evaluation.Project throws exceptions when building several times (not reproduced on our side). We send a test NuGet packaged with this fix and the user confirmed it works.

Fix:
Use a new ProjectCollection instance into the Project constructor (by default, it uses a global collection)

PR #756, opened 2015-12-20
Added missing ToISOString methods onto the Date and DateTime classes

maverix:
The current Date and DateTime classes are missing the ToISOString methods

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

UPD: #989


DaniilVeriga:
Thank you very much for the pull request. At the moment I am not sure if it is going to be merged, because we are about to split .NET and JavaScript API into different namespaces not to have them mixed altogether.

Anyways, before a possible merge of the pull request, please consider [signing a CLA form](https://docs.google.com/a/object.net/forms/d/12IAxRmWCk2vD-M3mj4d0ihztqKoiREpRRSVq4pEk0-k/viewform?c=0&w=1). 

There are related documents:
[Contributing document](https://github.com/bridgedotnet/Bridge/blob/master/CONTRIBUTING.md)
[CLA document](https://github.com/bridgedotnet/Bridge/blob/master/CLA.md)

maverix:
@DaniilVeriga No worries.  I'm glad that you are going to have a separation between .NET and Javascript.  Any hints how that is going to look?

DaniilVeriga:
Hi @maverix,

I am sorry for the delay. It will look like that. 

For example, `System.String` will have only .NET functionality. All the JavaScript String methods will go to, probably, `Bridge.String` or something like that. We are not sure about the namespace.

Similar with the DateTime class. All the native .NET DateTime functionality goes to `System.DateTime`. All JavaScript DateTime/Date will go to another namespace - Bridge.DateTime| Bridge.Date or Bridge.JavaScript.DateTime|Bridge.JavaScript.Date or Bridge.Ecmascript5.DateTime|Bridge.Ecmascript5.Date or something else. Again, we are not really sure yet on a namespace we'll stick to.

sdarnell:
I can easily see how that applies to types and static methods,  but what about instance methods? Presumably the JS variants will be declared via something like extension methods?

DaniilVeriga:
Agree, it looks like there is the only solution with that model - extension methods.

geoffreymcgill:
By the end of this week we should have a good idea on how this is going to work.

PR #774, opened 2015-12-22
Global ignore cast

Danielku15:
This is a small feature I added to Bridge to globally control the "IgnoreCast" setting. My project is not only targeting JavaScript and therefore I want to keep my core classes mostly clean from third party code (special attributes etc). 

I don't want to spoil my types with additional "IgnoreCast" attributes to control the setting I thought it would be nice to have it globally in the bridge.json.


DaniilVeriga:
@Danielku15, thank you very much for the pull request! We'll be reviewing.
Created a feature issue: #783

Issue #860, opened 2016-01-25
Can't build using MonoDevelop on Linux


MangelMaxime:
Hello guys,
I just installed Bridge using nuget packet but I can't build the project. If I use the VSCode project I can build it using the given task.

```

Building: bridge (Debug)

Build started 1/25/2016 10:26:37 PM.
__________________________________________________
Project "/home/shadaen/Projects/herebris/bridge/bridge.csproj" (Build target(s)):
	
	Target PrepareForBuild:
		Configuration: Debug Platform: AnyCPU
	
	Target CopyFilesMarkedCopyLocal:
		Copying file from '/home/shadaen/Projects/herebris/packages/Bridge.1.10.4/lib/net40/Bridge.dll' to '/home/shadaen/Projects/herebris/bridge/bin/Debug/Bridge.dll'
		Copying file from '/home/shadaen/Projects/herebris/packages/Bridge.1.10.4/lib/net40/Bridge.Html5.dll' to '/home/shadaen/Projects/herebris/bridge/bin/Debug/Bridge.Html5.dll'
		Copying file from '/home/shadaen/Projects/herebris/packages/Bridge.jQuery.2.8.0/lib/net40/Bridge.jQuery2.dll' to '/home/shadaen/Projects/herebris/bridge/bin/Debug/Bridge.jQuery2.dll'
	
	Target GenerateSatelliteAssemblies:
	No input files were specified for target GenerateSatelliteAssemblies, skipping.
	
	Target CoreCompile:
		Tool /usr/lib/mono/4.5/mcs.exe execution started with arguments: /noconfig /debug:full /debug+ /optimize- /out:obj/Debug/bridge.dll MyClass.cs Properties/AssemblyInfo.cs obj/Debug/.NETFramework,Version=v4.5.AssemblyAttribute.cs /target:library /define:DEBUG /nostdlib /reference:/usr/lib/mono/4.5/System.dll /reference:../packages/Bridge.1.10.4/lib/net40/Bridge.dll /reference:../packages/Bridge.1.10.4/lib/net40/Bridge.Html5.dll /reference:../packages/Bridge.jQuery.2.8.0/lib/net40/Bridge.jQuery2.dll /reference:/usr/lib/mono/4.5/mscorlib.dll /warn:4
CSC:  warning CS1685: The predefined type `System.Object' is defined multiple times. Using definition from `Bridge.dll'
[...]

Issue #531, opened 2015-10-24
Support .csproj wildcard configurations


geoffreymcgill:
http://forums.bridge.net/forum/bridge-net-pro/bugs/836

Issue #562, opened 2015-11-04
Support for Visual Studio shared projects


Ttxman:
Linking every shared cs (add existing item ... add as link) to more projects is very annoying. Xamarin and Microsoft solved this by adding "Shared Projects" == bunch of source files organized in project (*.shproj), that are automatically included at compile time.

https://developer.xamarin.com/guides/cross-platform/application_fundamentals/shared_projects/
http://blogs.msdn.com/b/dotnet/archive/2014/04/21/sharing-code-across-platforms.aspx

I made a little change in source Translator.ProjectReader.cs to support it locally, there is the code that works for me:
```C#
        protected virtual IList<string> GetSourceFiles(XDocument doc)
        {
            var result = new List<string>();

            var nodeList = from n in doc.Descendants()
                           where n.Name.LocalName == "Compile" &&
                                 !string.IsNullOrWhiteSpace(n.Attribute("Include").Value)
                           select n;

            foreach (var node in nodeList)
            {
                result.Add(node.Attribute("Include").Value);
            }


            var sharedProjects = doc.Descendants()
                .Where(n => n.Name.LocalName == "Import" && n.Attributes().Any(a => a.Name.LocalName == "Label" && a.Value == "Shared"))
                .Select(n => n.Attribute("Project").Value);


            foreach (var proj in sharedProjects)
            {
                string path = Path.Combine(Path.GetDirectoryName(Location), proj);
                doc = XDocument.Load(proj);
                
                nodeList = from n in doc.Descendants()
                           where n.Name.LocalName == "Compile" &&
                                 !string.IsNullOrWhiteSpace(n.Attribute("Include").Value)
                           select n;

                string rootpath = Path.GetDirectoryName(path)+"\\";

                foreach (var node in nodeList)
                {
                    result.Add(node.Attribute("Include").Value.Replace("$(MSBuildThisFileDirectory)", rootpath));
                }
            }
            return result;
        }

```

I didn't make this a pull request, because I'm using vs2015 and there are some modifications of *.csproj files to make the new nuget work and i don't want to install vs2013 and maintain 2 branches:)

Suchiman:
Related: http://forums.bridge.net/forum/general/feature-requests/878-support-shared-projects

vladsch:
Can you retest with latest code from 'release 1.10' branch? We changed how source files are extracted from project file, now we use MSBuild API, may be it automatically can resolve the issue

geoffreymcgill:
[...]

PR #880, opened 2016-01-30
[Issue860] Addresses a translator issue with Mono and Project reading.

fabriciomurta:
Microsoft.Builder.Evaluation.Project(string) overload has an issue in Mono
where it does not read the file at all.

The bug has been detailed in:
https://bugzilla.xamarin.com/show_bug.cgi?id=38224

As an approach that works on both Windows and Linux, the correct overload
for the project is chosen just by changing the call from:
Project(fileName)

to:
Project(XmlReader.Create(fileName))

Which is anyway the chosen overload when just the first form is chosen.

The issue in Bridge is triggered by
Bridge.Translator.ProjectReader.GetSourceFiles()
Called when Bridge builder is about to read the list of source files inside
the .csproj project file.

Current Mono version always returns an empty list of files without the fix
this commit introduces.

This issue has been introduced with #531, released on Bridge 1.10.0, and
reported on #860, which may be released to 1.11.0 or 1.10.5.


LeonidVeriga:
Exception details thrown when building **Bridge.sln** (with test projects enabled) in Windows:
```
9>C:\NET\BridgeNew\Bridge\Testing\tests\clientOne\Bridge.Builder.Console : error 2016-02-02T22: 35:56 Error: exception level: 0 - Value cannot be null.
9>  Parameter name: key
9>  Stack trace:
9>     at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
9>     at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
9>     at Microsoft.Build.Evaluation.ProjectCollection.LoadedProjectCollection.RemoveProject(String projectFullPath, Project project)
9>     at Microsoft.Build.Evaluation.ProjectCollection.UnloadProject(Project project)
9>     at Bridge.Translator.Translator.GetSourceFiles(XDocument doc) in C:\NET\BridgeNew\Bridge\Compiler\Translator\Translator\Translator.ProjectReader.cs:line 152
9>     at Bridge.Translator.Translator.ReadProjectFile() in C:\NET\BridgeNew\Bridge\Compiler\Translator\Translator\Translator.ProjectReader.cs:line 19
9>     at Bridge.Translator.Translator.Translate() in C:\NET\BridgeNew\Bridge\Compiler\Translator\Translator\Translator.cs:line 99
```

LeonidVeriga:
The current code after 3f9e07b 
 * fully works on Windows
 * Works on Mono but #531 broken (#562)

LeonidVeriga:
@fabricio, we need to check if the current changes in the code work on Mono and whether #531 broken or not (on Mono also). To check that, **Bridge.sln** should be compiler with the Testing projects loaded, the client tests run and then checking server-side NUnit tests (especially test 18).

Issue #918, opened 2016-02-07
Problem awaiting a dynamic variable


johnwason:
Awaiting a dynamic variable does not generate "taskResult".  For instance:

        dynamic a = new App() as dynamic;
        Console.WriteLine((int)await a.Test());

emits

        $task1.getResult();
        console.log(Bridge.cast($taskResult1, Bridge.Int));

when it should be

        $taskResult1=$task1.getResult();
        console.log(Bridge.cast($taskResult1, Bridge.Int));

http://live.bridge.net/#bef4a120fffb2fba2eaf

vladsch:
Compiler doesn't know return type of Test method because it is dynamic method. So, you need to cast Test invocation, also Bridge doesn't change case for dynamic code (Test will be emitted as is, like "Test", not "test")

So, you shoud use
`Console.WriteLine((int)await (Task<int>)a.test());`

johnwason:
Would it be possible to implicitly detect when a return value is expected?  This works in standard C# without a cast and I think it will confuse the users of my library to have to cast before the await.

vladsch:
yes, we will try to fix it. Thanks for your test case

vladsch:
It will be fixed in 1.11
Just please note that Bridge doesn't change case of dynamic code because compiler doesn't know about type of dynamic expression (it can be type reference or method)
So, if you write dynamic code then you should pay attention on case of the code

For example, `await a.Test()` will not work because real `Test` method is emitted as `test`. Therefore you should use lower case also in dynamic expression `await a.test()`

vladsch:
Also, using dynamic awaiter is not good idea because we have no type information during compilation. So, we cannot get information about how to get awaiter (for dynamic code we assume that it is Task). But awaitable object can be any object which has GetAwaiter method (or extension method). In this case, Bridge will not able to get info about such methods. 
It is better if Bridge will throw an exception if dynamic code is used as awaiter (not in v1 but in v2) because it is easy to get incorrect javascript

johnwason:
This test case fails with #924 but works with #925:

    public class App
    {
        public async Task<int> test()
        {
            return 1;
        }
    

        [Ready]
        public static async void Main()
        {
            dynamic a = new App() as dynamic;
            await a.test();
            var c = await a.test();
            Console.WriteLine((int)await a.test());
            App b = new App();
            await b.test();            
        }
    }

johnwason:
[...]

PR #925, opened 2016-02-08
Return value for await operation on dynamic variable fix

johnwason:
This is a simple patch to fix Issue #918 .  It works with my test cases but does not check the return type. There may need to be a Bridge.cast added to make sure the return type is the expected type. This may or may not be necessary depending on the level of strictness Bridge.net expects.


vladsch:
Thanks for your pull request. The issue is already fixed (almost the same code)
https://github.com/bridgedotnet/Bridge/pull/924

Issue #938, opened 2016-02-12
"The use of a keyword is invalid" in Edge browser


johnwason:
I am getting this error when testing in the Edge browser:

    SCRIPT1048: The use of a keyword for an identifier is invalid
    bridge.js (6983,31)

The code at that line is:

       _getResult: function (await) {
            switch (this.status) {
                case Bridge.TaskStatus.ranToCompletion:
                    return this.result;
                case Bridge.TaskStatus.canceled:
                    var ex = new Bridge.TaskCanceledException(null, this);
                    throw await ? ex : new Bridge.AggregateException(null, [ex]);
                case Bridge.TaskStatus.faulted:
                    throw await ? (this.exception.innerExceptions.getCount() > 0 ? this.exception.innerExceptions.get(0) : null) : this.exception;
                default:
                    throw new Bridge.InvalidOperationException("Task is not yet completed.");
            }
        },

So far my tests are running properly in Chrome, Firefox, IE, and Safari.

geoffreymcgill:
Hi @johnwason: Is there are code sample we can run to reproduce? or is it just being thrown when **bridge.js** is loaded?

geoffreymcgill:
Pull Request: https://github.com/bridgedotnet/Bridge/pull/940

PR #940, opened 2016-02-12
[#938] "The use of a keyword is invalid" in Edge browser

geoffreymcgill:


LeonidVeriga:
The issue #938 has been fixed by the change.
However, Edge gives 9 test errors for Issue549.

Issue #1006, opened 2016-02-28
Allow optional formatProvider argument on toBoolean and other Convert js functions


geoffreymcgill:
### Expected behaviour

```javascript
var flag = Bridge.Convert.toBoolean("true");
```

### Actual behaviour

```javascript
var flag = Bridge.Convert.toBoolean("true", null);
```

### Steps to reproduce the behaviour

```csharp
var flag = Convert.ToBoolean("true");
```

PR #1007, opened 2016-02-28
Allow optional formatProvider argument on toBoolean and other Convert js functions

geoffreymcgill:
Some formatting of the files was also done


geoffreymcgill:
Related to #1006

Issue #1080, opened 2016-03-19
Support for C# Random class


geoffreymcgill:
[MSDN Random](// https://msdn.microsoft.com/en-us/library/system.random(v=vs.110).aspx)

### Steps To Reproduce

```csharp
public class App
{
    [Ready]
    public static void Main()
    {
        Random r = new Random();

        Console.WriteLine(r.Next(10));
        Console.WriteLine(r.Next(10));
        Console.WriteLine(r.Next(10));

        Console.WriteLine(r.Next(5, 10));
        Console.WriteLine(r.Next(5, 10));
        Console.WriteLine(r.Next(5, 10));

        Console.WriteLine(r.NextDouble());
        Console.WriteLine(r.NextDouble());
        Console.WriteLine(r.NextDouble());

        byte[] buffer = new byte[5];

        r.NextBytes(buffer);
        Console.WriteLine(string.Join(",", buffer));

        r.NextBytes(buffer);
        Console.WriteLine(string.Join(",", buffer));
    }
}
```

michaelcheers:
@geoffreymcgill 
I've made a pull request for this: [Pull](https://github.com/bridgedotnet/Bridge/pull/1118)

LeonidVeriga:
Implementation details:

What we did is:
 - Take **.Net** `Random` class;
 - Transpile it using **Bridge.NET**.

Basically, there are four files:
- [MS Random C# source file](http://referencesource.microsoft.com/#mscorlib/system/random.cs), formatted just a little [Random.cs](https://github.com/bridgedotnet/Bridge/blob/Random/Transpiled/System/Random.cs) 
- [Javascript file](https://github.com/bridgedotnet/Bridge/blob/Random/Bridge/Resources/random.js) generated by **Bridge.NET** from [Random.cs](https://github.com/bridgedotnet/Bridge/blob/Random/Transpiled/System/Random.cs) -
  just 141 lines of code, 100% fits the .Net `Random` class functionality
- **Bridge** [Random.cs](https://github.com/bridgedotnet/Bridge/blob/Random/Bridge/System/Random.cs) file with `[External]` attribute applied, some methods have `[Template]` to control their overrides
- [Tests](https://github.com/bridgedotnet/Bridge/blob/Random/Tests/Batch1/RandomTests.cs) taken from [MS corefx tests](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/Random.cs) slightly formatted to have text description in assertions.

PR #1118, opened 2016-03-29
Random Support

michaelcheers:
Addresses #1080


LeonidVeriga:
Hi @michaelcheers!

Thank you for your time and contribution implementing `Random` class!

We reviewed the code and found it very good. It has big chunk of functionality on js side (that is almost 1k lines of code). This functionality looks a bit redundant compared to .Net's one. And it got us thinking of trying the "native" .Net implementation.

What we did is:
 - Take **.Net** `Random` class;
 - Transpile it using **Bridge.NET**.

Basically, there are four files:
- [MS Random C# source file](http://referencesource.microsoft.com/#mscorlib/system/random.cs), formatted just a little [Random.cs](https://github.com/bridgedotnet/Bridge/blob/Random/Transpiled/System/Random.cs) 
- [Javascript file](https://github.com/bridgedotnet/Bridge/blob/Random/Bridge/Resources/random.js) generated by **Bridge.NET** from [Random.cs](https://github.com/bridgedotnet/Bridge/blob/Random/Transpiled/System/Random.cs) -
  just 141 lines of code, 100% fits the .Net `Random` class functionality
- **Bridge** [Random.cs](https://github.com/bridgedotnet/Bridge/blob/Random/Bridge/System/Random.cs) file with `[External]` attribute applied, some methods have `[Template]` to control their overrides
- [Tests](https://github.com/bridgedotnet/Bridge/blob/Random/Tests/Batch1/RandomTests.cs) taken from [MS corefx tests](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/Random.cs) slightly formatted to have text description in assertions.

We had an internal discussion and decided to use the approach above as it gets 100% **.Net** functionality with just 141 lines of javascript code.

Unfortunately we are not merging this Pull Request. However it pushed us to verify whether this nice approach of transpiling MS corefx code to javascript works for `Random` class.

Thanks again for contribution.

PR #1182, opened 2016-04-11
Issue1180 Structs as generic arguments are treated like classes

### LINKED ISSUE #1180: Structs as generic arguments are treated like classes
If you use a struct as a generic argument, it will be passed by reference everywhere and the $clone method will not be inserted anywhere.

### Expected

0
5

### Actual

5
5

### Steps To Reproduce

[Live](http://live.bridge.net/#f01f1e291e860821a6ab1b0913c888c7)

```csharp
public struct Vector2
{
    public float x;
    public float y;
}

public class App
{
    [Ready]
    public static void Main()
    {
        List<Vector2> list = new List<Vector2>();
        list.Add(new Vector2() { x = 0.0f, y = 1.0f });
        
        Vector2 vec = list[0];
        
        vec.x = 5.0f;
        
        Window.Alert(list[0].x.ToString()); // Oops.
        Window.Alert(vec.x.ToString());
    }
}
```

vladsch:
Fixes #1180.


Tests/Runner/Batch1/code.js
```diff
@@ -3103,7 +3103,7 @@ SomeExternalNamespace.SomeNonBridgeClass.prototype.foo = function(){return 1;};
                     x: 5,
                     y: 7
                 } );
-                var c = b.test2(a.$clone());
+                var c = b.test2(a.$clone()).$clone();
```
LeonidVeriga: @vladsch, this change does not look as expected

Tests/Runner/Batch1/code.js
```diff
@@ -6272,7 +6318,7 @@ SomeExternalNamespace.SomeNonBridgeClass.prototype.foo = function(){return 1;};
                 var a = Bridge.ClientTest.BridgeIssues.Bridge577.someMethodA(1);
                 Bridge.Test.Assert.notNull$1(a, "#577 Bridge577UnitA created");
     
-                var b = Bridge.ClientTest.BridgeIssues.Bridge577.someMethodB(7);
+                var b = Bridge.ClientTest.BridgeIssues.Bridge577.someMethodB(7).$clone();
```
LeonidVeriga: @vladsch, this change does not look as expected

Tests/Runner/Batch1/code.js
```diff
@@ -8629,7 +8675,7 @@ SomeExternalNamespace.SomeNonBridgeClass.prototype.foo = function(){return 1;};
     
                 var value1 = Bridge.Nullable.getValueOrDefault(test1, 0);
                 var value2 = Bridge.Nullable.getValueOrDefault(test2, new Bridge.ClientTest.BridgeIssues.Bridge762A());
-                var value3 = Bridge.Nullable.getValueOrDefault(test3, new Bridge.ClientTest.BridgeIssues.Bridge762B());
+                var value3 = Bridge.Nullable.getValueOrDefault(test3, new Bridge.ClientTest.BridgeIssues.Bridge762B()).$clone();
```
LeonidVeriga: @vladsch, this change does not look as expected

Tests/Runner/Batch1/code.js
vladsch: i guess it looks good, we should clone struct when we assign to new variable

Tests/Runner/Batch1/code.js
vladsch: i guess it looks good, we should clone struct when we assign to new variable

PR #1214, opened 2016-04-17
Issue1200 Adding handler to event in generic class silently fails

### LINKED ISSUE #1199: Adding handler to event in generic class silently fails
### Expected
Adding handlers to event exposed by generic class should register that handler within event. 

### Actual
Event is not registered - it silently fails without any Javascript exception. 
Problem lays with naming after translation to JS. Workaround is to make event name follow camel case convention. Pull request will follow.

[Live Bridge](http://live.bridge.net/#6f8e544542dfd7797698a27245fac462)

### Steps To Reproduce

```csharp
public class IssueEvents
    {
        public delegate string SomeDel();

        class NotWorking<T>
        {
            public event SomeDel IsNotWorking;
            public string Validate() {
                return IsNotWorking == null ? "no subscribers": IsNotWorking();
            }
        }
        
        [Ready]
        public static void Main()
        {
            var wrong = new NotWorking<int>();
            wrong.IsNotWorking += () => "somevalue";
            Window.Alert(wrong.Validate()); // in browser gives "no subscribers" instead of "somevalue"
        }
    }
```

vladsch:
Fixes #1199. (commit code uses incorrect 1200 number for issue)


Compiler/Translator/Emitter/Emitter.Helpers.cs
```diff
@@ -433,7 +433,7 @@ public virtual string GetEntityName(IEntity member, bool forcePreserveMemberCase
             }
 
 
-            if (member is IMember && this.IsMemberConst((IMember)member)/* || member.DeclaringType.Kind == TypeKind.Anonymous*/)
+            if (member is IMember && this.IsMemberConst((IMember)member) || member is IEvent)
             {
                 preserveMemberChange = true;
             }
```
LeonidVeriga: #1199

Tests/Batch1/BridgeIssues/1200/N1200.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        [Test]
+        public static void TestIEvent()
+        {
+            var wrong = new NotWorking<int>();
+            wrong.IsNotWorking += () => "somevalue";
+            Assert.AreEqual("somevalue", wrong.Validate());
+        }
+    }
+}
```
LeonidVeriga: #1199

Issue #1234, opened 2016-04-19
Use bridge.min.js in demo.html.pp template


geoffreymcgill:
### Expected

```html
<script src="../output/bridge.min.js"></script>
```

### Actual

```html
<script src="../output/bridge.js"></script>
```

### Steps To Reproduce

Browse to **/Bridge/www/** folder and view source of **demo.html** file.

PR #1235, opened 2016-04-19
[#1234] Use bridge.min.js in demo.html.pp template

### LINKED ISSUE #1244: Use bridge.min.js in demo.html.pp template
### Expected

**demo.html.pp** template uses** bridge.min.js**

### Actual

**demo.html.pp** template uses** bridge.js**

If the debug version required, just remove `.min` from the path in **demo.html**

Using the `.min` source in **demo.html** also provides a hint to developers than an actual min version is available. It would appear many are missing this feature.

This change only applies to the auto generated demo.html file which is created as a sample to help get developers started quickly.

geoffreymcgill:
Fixes #1244

### DIFF
--- a/.build/files/demo.html.pp
+++ b/.build/files/demo.html.pp
@@ -4,13 +4,12 @@
 <head>
     <meta charset="utf-8" />
     <title>Bridge $rootnamespace$</title>
-    <script src="../output/bridge.js"></script>
+    <script src="../output/bridge.min.js"></script>
     <script src="../output/$rootnamespace$.js"></script>
 </head>
 <body>
     <!-- 
-        Right-Click on this file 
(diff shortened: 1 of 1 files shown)


michaelcheers:
I don't think this is a good idea as I like to be able to debug bridge.js, of course: I can just change it.
Anyway, you can do it if you like.

geoffreymcgill:
If you need the debug version, just remove `.min` from the path in demo.html.

Using the `.min` source in demo.html also provides a hint to developers than an actual min version is available. It would appear many are missing this feature.

Of course, in your actual application .html files, use whichever .js file is appropriate. This change only applies to the auto generated demo.html file which is created as a sample to help get developers started quickly.

PR #1265, opened 2016-04-27
Make stock object's GetHashCode() repeatable for classes with fields

### LINKED ISSUE #1264: Default GetHashCode() is not repeatable
### Expected
Default GetHashCode() invocations for classes having fields being custom classes should return same values for unchanged object.

### Actual
Actually it returns different values.

### Steps To Reproduce

**UPD**: Example ([Live](http://live.bridge.net/#4d4e1fd1122f7ae823237fa44cbf174c)):
```csharp

	public class Bar {}

	class Foo {
		private object _bck = new Bar();
	}

	public class App
	{
		[Ready]
		public static void Main()
		{	
			var foo = new Foo();
			var h1 = foo.GetHashCode();
			var h2 = foo.GetHashCode();

			Console.WriteLine(h1 != h2 ? "has bug" : "no bug");
		}
	}
```

d-p-y:
Fixes #1264

- Core.js behavior when fields are detected

### DIFF
--- a/Bridge/Resources/Core.js
+++ b/Bridge/Resources/Core.js
@@ -214,7 +214,7 @@
             }
         },
 
-        getHashCode: function (value, safe, store) {
+        getHashCode: function (value, safe, store = true) {
             if (Bridge.isEmpty(value, true)) {
                 if (safe) {
                     return 0;
@@ -299,7 +299,7 @@
                     delete Bridge.$$hashCodeCache;
                 }
 
-                if (result !== 0) {
+                if (result !== 0 && !store) {
                     return result;
                 }
             }
--- a/Bridge/Resources/Core.js
+++ b/Bridge/Resources/Core.js
@@ -214,7 +214,14 @@
             }
         },
 
-        getHashCode: function (value, safe, store = true) {
+        getHashCode: function (value, safe, store) {
+            //in CLR: mutable object should keep on returning same value
+            //bridge.net goals: make it deterministic (to make testing easier) without breaking CLR contracts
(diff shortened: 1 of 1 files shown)


DaniilVeriga:
Thank you for the pull request! We will review.

LeonidVeriga:
Hi @d-p-y,

Thank you for the Pull Request!

And yes, unfortunately, it breaks a couple of tests. Could we try this approach - I created a [Live Example](http://live.bridge.net/#3e194168d80eb3794f8853dc8a94e521) for the tests that fail so that you could check the fix.

One note from my side, default parameters are [not supported](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Default_parameters) in IE. I would suggest using an alternative construction in a function body like:
```
var b = typeof b !== 'undefined' ?  b : 1;
```

d-p-y:
Leonid,
I've changed implementation - it doesn't break tests 341,467, 470 anymore. It does it by implementing additional requirements for GetHashCode() as needed for Bridge.net tests.
I suggest to give function a comment as I did in my updated PR. Otherwise it is difficult to know its extra goals that are not present in CLR specification. 
Hope it is useful now.

Kind regards
Dominik

LeonidVeriga:
@d-p-y,

Yes, the code works just perfect; merged to `master` for `1.13` release.

Just added a [test](https://github.com/bridgedotnet/Bridge/commit/2b4806f3caee2cbf513a5fd2deb0ac3d861a689a)

Thank you!

PR #1333, opened 2016-05-10
Issue1328 Incorrect setting-to-null of optional constructor argument when using named arguments 

### LINKED ISSUE #1328: Incorrect setting-to-null of optional constructor argument when using named arguments
Related forum thread:
http://forums.bridge.net/forum/bridge-net-pro/bugs/2324-incorrect-setting-to-null-of-optional-constructor-argument-when-using-named-arguments

### Expected

```js
var l2 = new Demo.Link2("url", "test");
// or,
var l2 = new Demo.Link2("url", "test", void 0);
```

### Actual

```js
var l2 = new Demo.Link2("url", "test", null);
```

### Steps To Reproduce

http://live.bridge.net/#9709541e32e342b3697250c83ba7d408

```csharp
var l2 = new Link2(url: "url", text: "test");

public  class Link2
{
    public Link2(string url, string text, Optional<string> name = new Optional<string>())
    {
    }
}
```

vladsch:
Fixes #1328.


LeonidVeriga:
The issue is basically fixed; waiting for #1340 fix to pass the tests and check if more fixing required

PR #1369, opened 2016-05-16
Fixed #156 and #1374, fixed #1311, formatted Bridge code.

### LINKED ISSUE #156: 1D array does not throw exception when referenced outside its boundaries
For C#, when an array is touched outside its allocated area, and exception is thrown, as it could be touching any unexpected memory address.

For JavaScript, this is not true. Accessing an array beyond its boundaries just returns 'null' if that's actually beyond arrays' defined data. Also returns null when the element has not been set (non-contiguous arrays).

```c#
namespace ClassLibrary1
{
    public partial class Class1
    {
        public static int[] MyArray = new int[5];
        public static int[,] My2DArray = new int[5, 5];

        public static void TriggerError()
        {
            Console.WriteLine("Array value outside its bounds is: " + MyArray[10]);
            Console.WriteLine("Next command will trigger an exception, whereas it should just show 'null'.");
            Console.WriteLine("Breaking array outside its bounds is: " + My2DArray[2, 10]);
        }
    }
}
```

Will result in the JS code:

```js
Bridge.define('ClassLibrary1.Class1', {
    statics: {
        config: {
            init: function () {
                this.myArray = new Array(5);
                this.my2DArray = Bridge.Array.create(0, null, 5, 5);
                Bridge.ready(this.triggerError);
            }
        },
        triggerError: function () {
            console.log("Array value outside its bounds is: " + ClassLibrary1.Class1.myArray[10]);
            console.log("Next command will trigger an exception, whereas it should just show 'null'.");
            console.log("Breaking array outside its bounds is: " + ClassLibrary1.Class1.my2DArray.get([2, 10]));
        }
    }
});
```

And an exception will be triggered once the third command on `triggerError` function is executed on the browser. The exception is thrown from Bridge.Array.get(), corresponding to:

```
Bridge.js:4275                     if (indices[i] < 0 || indices[i] >= arr.$s[i]) {
Bridge.js:4276                        throw new Bridge.ArgumentException("Index " + i + " out of range");
BRidge.js:4277                    }
```

A similar exception should be thrown for the myArray invalid access on the code above.

## See Also

* #2466

### LINKED ISSUE #1311: Enum parsing of number doesn't work
### Expected

No error

### Actual

Invalid Enumeration Value

### Steps To Reproduce
[Bridge Fiddle](http://live.bridge.net/#22ee543e8dda2fe46ba8a2e3388d4864)
[Dot Net Fiddle](https://dotnetfiddle.net/azS3Cp)
```csharp
public static class Program {
    public enum SimpleEnum {
        A,
        B = 3,
        C,
        D = 10
    }

    public enum ByteEnum : byte {
        A,
        B = 3,
    }
[Ready]
    public static void Main (string[] args) {
        Console.WriteLine(string.Format("{0} {1} {2}", SimpleEnum.A, SimpleEnum.B, SimpleEnum.C));
        Console.WriteLine(string.Format("{0} {1}", ByteEnum.A, ByteEnum.B));
        Console.WriteLine(string.Format("{0} {1} {2}", (int)SimpleEnum.A, (int)SimpleEnum.B, (int)SimpleEnum.C));
        Console.WriteLine(string.Format("{0} {1}", (int)ByteEnum.A, (int)ByteEnum.B));
        Console.WriteLine(string.Format("{0} {1}", Enum.Parse(typeof(SimpleEnum), "C"), Enum.Parse(typeof(SimpleEnum), "3")));
    }
}
```

michaelcheers:
#1404 Code clean up with extern
#1374 Use map function for Array.ConvertAll


michaelcheers:
I couldn't seem to get the unit tests to work.

LeonidVeriga:
@michaelcheers, thank for the pull request.

We will review the changes proposed and will try to include it in the upcoming releases.

Regarding unit tests, our apologies for the inconvenience. At the moment **Bridge.Test** is not publicly available. For now the only option is to unload everything within the *Tests* folder from the Solution. We are getting closer with releasing it. We'll tweet once it gets available.

michaelcheers:
I've just formatted Bridge a bit with the last commit.

Bridge/System/Action.cs
```diff
@@ -3,78 +3,78 @@
...
     [Name("Function")]
     [IgnoreGeneric]
-    [IgnoreCast]
+    [IgnoreCast, External]
     public delegate void Action<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12>(T1 arg1, T2 arg2, T3 arg3, T4 arg4, T5 arg5, T6 arg6, T7 arg7, T8 arg8, T9 arg9, T10 arg10, T11 arg11, T12 arg12);
 
     [Name("Function")]
     [IgnoreGeneric]
-    [IgnoreCast]
+    [IgnoreCast, External]
```
geoffreymcgill: Please separate into individual Attributes. One per line.

```c#
// Existing
[IgnoreCast, External]

// Revised
[IgnoreCast]
[External]
```

Bridge/Resources/Enum.js
```diff
@@ -24,6 +24,9 @@
         },
 
         parse: function (enumType, s, ignoreCase, silent) {
+            var intReturnValue = {};
+            if (Bridge.Int32.tryParse(s, intReturnValue))
```
geoffreymcgill: Wrapping curly braces are required to maintain a consistent coding style throughout the project. 

```js
// Existing
if (Bridge.Int32.tryParse(s, intReturnValue))
    return intReturnValue.v;

// Revised
if (Bridge.Int32.tryParse(s, intReturnValue)) {
    return intReturnValue.v;
}
```

michaelcheers: @geoffreymcgill, I'll change it.

Bridge/System/Action.cs
```diff
@@ -3,78 +3,86 @@
 namespace System
 {
     [Name("Function")]
-    [IgnoreCast, IgnoreGeneric]
+    [IgnoreCast, IgnoreGeneric, External]
```
michaelcheers: @geoffreymcgill, these were already joint so I just followed the style. Do you want me to change those too?

geoffreymcgill: Sorry about that. Now I see they were not separated originally. 

If you're in the files, and don't mind separating, that would be great.

michaelcheers: @geoffreymcgill, I'll separate them.

michaelcheers:
@LeonidVeriga, I think my [first commit](https://github.com/bridgedotnet/Bridge/pull/1369/commits/0ff7b346294b133c0932389bfa645b0d80015c46) was a bad one, let me explain:
A one dimensional array doesn't have a definition for $.get().
For example this code:
```csharp
Console.WriteLine(new[]{1,2,3}[0]);
```
compiles to:
```javascript
console.log([1,2,3].$get([0]));
```
and the javascript doesn't actually work because normal 1D arrays (which are javascript arrays) don't have a definition for $get.
So, either we make a definition on the prototype of Array or we use the definition Bridge.Array.getItem(array, index).

LeonidVeriga:
@michaelcheers, we have reviewed the pull request and partially merged the changes proposed.
Thank you for your time and contribution!

#1311 fix and `extern` changes are really great.

You can review the commits we made in the process of merging this pull request (they all linked to this PR). Here is a summary:
- #1311 fix merged with no changes
- for `extern` changes created an issue #1404 for tracking purposes, merged with some changes
  However, there is some [specific](https://github.com/bridgedotnet/Bridge/issues/1404#issuecomment-222301402) related to C#6
- [#1374](https://github.com/bridgedotnet/Bridge/issues/1374#issuecomment-222277194) fix changed: using `map()` instead of `convertAll()` is not recommended as the latter contains code to check if input `array` or `converter` is `null`with exception throwing
- #156 fix reverted as it brakes several tests. Yes, **Bridge.Test** is not released so I narrowed down a couple of tests that, in addition to your own checks, you can use to verify a fix - [Live](http://live.bridge.net/#2baeb58d051f76e7da7ff18b455d6f9d)
- changing `static readonly` to `const` reverted. However, it forced me to review tests and added a commit to improve the tests.

Creating pull requests, we tries to follow the rules:
- adding a description of changes - what it addresses and how it fixes. This is helpful and speeds up merging process;
- when possible, using latest Bridge source code as a pull request branch base. This minifies merge commits;
- one Pull Request per Issue in separate branch (unless issues are related to each other like [here](https://github.com/bridgedotnet/Bridge/pull/1367#issue-154926398));
  This **really** helps in merging pull requests - the smaller the changes the easier to manage, track and merge. Especially given the changes should be verified and tested carefully, they may be postponed or reverted, there are other pull requests from the team and the Bridge community being merged potentially causing merge conflicts.

michaelcheers:
@LeonidVeriga, See my new pull request: #1407.

PR #1371, opened 2016-05-17
Cast failed for svg tag from Node to Element on iOS WKWebView

liruqi:
A svg Node with `NodeType = 1`, cast to Element failed, Bridge.as may fail on iOS WKWebView.
No need to throw exception in this case.
BTW, in my understanding, `cast` is something we are trying to do. We should not throw exception for cast fail. Let exception happen somewhere else look good to me.

### DIFF
--- a/Bridge/Resources/Core.js
+++ b/Bridge/Resources/Core.js
@@ -364,6 +364,9 @@
             var result = Bridge.as(obj, type, allowNull);
 
 	        if (result === null) {
+	            if (navigator.platform && navigator.platform.substr(0,2) === 'iP') {
+	                return obj;
+	            }
 	            throw new Bridge.InvalidCastException("Unable to cast type " + (obj ? Bridge.getTypeName(obj) : "'null'") + " to type " + Bridge.getTypeName(type));
 	        }
 


LeonidVeriga:
Hi @liruqi,

Thank you for the pull request!

Before a review of the pull request we would need the CLA [form](http://goo.gl/forms/CvDFLA3EPO) accepted. You can find more information in [CONTRIBUTING.md](https://github.com/bridgedotnet/Bridge/blob/master/.github/CONTRIBUTING.md). 

The full CLA is available [here](http://bridge.net/cla/).

When you get a chance, please review the CLA and complete the [CLA Acceptance Form](http://goo.gl/forms/CvDFLA3EPO). If you have any questions, please just let us know.

LeonidVeriga:
@liruqi, can you please provide more details on the problem the pull request is going to fix.
Having a [Live](http://live.bridge.net/) example would be great.

liruqi:
@LeonidVeriga I've submitted CLA form.
I got a live example using compiled bridge.js here: http://liruqi.github.io/Bridge/test/Bridge-1371.html

The problem is clear now:
Uncaught Unable to cast type SVGSVGElement to type HTMLElement

Any C# class representing SVGSVGElement?

michaelcheers:
@liruqi, The C# error message is: "Specified cast is not valid.". The as operator is for if you don't want it to throw an exception and will return null if the cast fails. You can use the As<Node>().

liruqi:
@michaelcheers I use cast in C# like this: `((Element) node)`
How to make the compiler translate it to `Bridge.as` instead of `Bridge.cast`?

liruqi:
Looks like I should do this:
`if (node.NodeType == 1) { Element e = node as Element; }`

geoffreymcgill:
Or, 

```c#
if (node.NodeType == 1) 
{ 
    Element e = node.As<Element>(); 
}
```

liruqi:
@geoffreymcgill Any plan to implement `SVGSVGElement`?

geoffreymcgill:
@liruqi No plans at the moment, but I can create a feature request and tag as `up-for-grabs`.

geoffreymcgill:
Issue created, see #1372.

liruqi:
Root cause for this issue: bridge.net translate C# `Element` to `HTMLElement`. 
Should translate to `Element` instead.

http://live.bridge.net/#4b100ceeca89eec5e764983ff1cae7b6

geoffreymcgill:
@liruqi - Thanks for the pull-request and comments. We're scrambling to finish up the next release (1.14) and haven't had the opportunity to fully review. 

At first look, your change looks correct, and for some reason the inheritance was incorrectly configured originally.

LeonidVeriga:
Hi @liruqi,

In 1.14 we reviewed HTML elements  (#1456, #1459) and `Element` is translated into `Element` ([Live](http://live.bridge.net/#7c6d05f8a91229fd53e0bb9999f2ac17)).

PR #1384, opened 2016-05-22
Fixed #1339.

### LINKED ISSUE #1339: Accessing to constants from derived class fails
### Expected

Class.SomeConstant (where Class is derived from BaseClass and SomeConstant is declared in BaseClass) should gives a value.

### Actual

It gives 'undefined'

### Steps To Reproduce

Example(http://live.bridge.net/#aa29f123765eb116b226c9bccc547aa4):

```csharp
public class FooBase
{
    public const string Bar = "Str";
}

public class Foo:FooBase
{}

/// <summary>
/// Creates a simple class
/// </summary>
public class App
{
    [Ready]
    public static void Main()
    {
        switch("Str")
        {
            case Foo.Bar: Global.Alert("Success"); break;
            default: Global.Alert("Failed"); break;
        }
    }
}
```

michaelcheers:


Bridge/Resources/Class.js
```diff
@@ -290,6 +290,12 @@
 
             prototype.$$name = className;
 
+            for (let x of inherits) {
```
geoffreymcgill: The `let` statement is not supported in IE9 + IE10, which for the time being are browsers we support with Bridge.

Can you modify to use `var`?

michaelcheers: @geoffreymcgill, I will do it in about 5 minutes.

LeonidVeriga:
Closing as the issue has been already fixed.

PR #1403, opened 2016-05-27
Issue1402 Byte conversions from long don't work

### LINKED ISSUE #1402: Byte conversions from long don't work
### Expected

255

### Actual

0

### Steps To Reproduce
[Dot Net Fiddle](https://dotnetfiddle.net/znHEoP)
[Live Bridge](http://live.bridge.net/#da74541dddf449cc104ac314f46c868b)
```csharp
/// <summary>
/// Creates a simple class
/// </summary>
public class App
{
    [Ready]
    public static void Main()
    {
        Console.WriteLine(GetBytes(long.MaxValue));
    }
    
    public static byte GetBytes (long value)
    {
        return unchecked((byte)(value >> 2));
    }
}
```

vladsch:
Fixes #1402.


Tests/Batch1/BridgeIssues/1400/N1402.cs
```diff
@@ -0,0 +1,21 @@
...
+        [Test]
+        public static void TestLongClipping()
+        {
+            long value = long.MaxValue;
+            Assert.AreEqual(255, (byte)(value >> 2));
+            Assert.AreEqual(-1, (sbyte)(value >> 2));
+            Assert.AreEqual(-1, (short)(value >> 2));
+            Assert.AreEqual(65535, (ushort)(value >> 2));
+            Assert.AreEqual(-1, (int)(value >> 2));
+            Assert.AreEqual(4294967295, (uint)(value >> 2));
```
michaelcheers: @vladsch, why not change 255 to 0xff, 65535 to 0xff, 4294967295 to 0xffffffff to make the meaning more clear?

LeonidVeriga: @michaelcheers, on the one hand it makes sense on the other hand it is the same thing for almost all developers) I'll change it to hex presentation.

PR #1407, opened 2016-05-28
Fixed #156 and general array behavior problems. See #1369.

### LINKED ISSUE #156: 1D array does not throw exception when referenced outside its boundaries

michaelcheers:
[In the last comment of #1369](https://github.com/bridgedotnet/Bridge/pull/1369/#issuecomment-222292597), I mentioned this.
It broke some of the arrays so I made this pull request which shouldn't do so badly for #156.

### DIFF
--- a/Bridge/Resources/Array.js
+++ b/Bridge/Resources/Array.js
@@ -310,7 +310,10 @@ var array = {
 
     getItem: function (obj, idx) {
         if (Bridge.isArray(obj)) {
-            return obj[idx];
+            if (idx < obj.length && idx > -1) {
+                return obj[idx];
+            }
+            throw new Bridge.ArgumentException("Index " + idx + " out of range");
         } else if (Bridge.isFunction(obj.get)) {
             return obj.get(idx);
         } else if (Bridge.isFunction(obj.getItem)) {
--- a/Compiler/Translator/Emitter/Blocks/IndexerBlock.cs
+++ b/Compiler/Translator/Emitter/Blocks/IndexerBlock.cs
@@ -1067,7 +1067,12 @@ namespace Bridge.Translator
 
         protected virtual void EmitSingleDimArrayIndexer(IndexerExpression indexerExpression)
         {
-            this.Write("System.Array.getItem(");
+            var index = indexerExpression.Arguments.First();
+            var primitive = index as PrimitiveExpression;
+            var cond = primitive != null && primitive.Value != null &&
+                Regex.Match(primitive.Value.ToString(), "^[_$a-z][_$a-z0-9]*$", RegexOptions.IgnoreCase).Success;
+            if (!cond)
+                this.Write("System.Array.getItem(");
             var oldIsAssignment = this.Emitter.IsAssignment;
             var oldUnary = this.Emitter.IsUnaryAccessor;
             this.Emitter.IsAssignment = false;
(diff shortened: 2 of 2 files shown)


LeonidVeriga:
Great, thanks @michaelcheers, we will review the pull request.

LeonidVeriga:
@michaelcheers,

That is a good idea to use `System.Array.getItem` (please note the current Bridge code uses `System.Array` instead of `Bridge.Array`) in `EmitSingleDimArrayIndexer` method (but a trade-off is checking index vs performance). However, there are two problems with the fix:
-  it misses closing parentthesis as it added `this.Write("System.Array.getItem(");` but did not added extra `this.WriteCloseParentheses()`;
-  `arr[0] = "x";` -> ` System.Array.getItem(arr,  0) = "x";` causing error JS `Invalid left-hand side in assignment`

michaelcheers:
@LeonidVeriga, I think you should make an option to use javascript indexer (s) instead of C# ones.

michaelcheers:
@LeonidVeriga, [I did add a close parentheses](https://github.com/bridgedotnet/Bridge/pull/1407/files#diff-f417604a9c2ff661009df53a4e986d9fL1122). I'll change it to System.Array and setItem (for setItem only if I can figure out how to).

LeonidVeriga:
@michaelcheers, regarding closing parentheses, yes, there is a call added ` this.WriteCloseParentheses();` but it is in a conditional `if` however the opening one ` this.Write("Bridge.Array.getItem(");` added unconditionally.

LeonidVeriga:
Thanks @michaelcheers. We will review the changes and preliminary it looks good.
1.14 release scope is defined already; we will have an internal discussion on 1.14 release date early next week. If we have time we include this pull request into 1.14.

michaelcheers:
@LeonidVeriga, I couldn't figure how to fix the setter problem.

kcudnik:
It would be nice when config is set to  "overflowMode": "Javascript", then array behaviour should be as in javascritp, and not like in c#, this will have better performance for javascript code since no getItem() that checks index range will be required

michaelcheers:
[Me too](https://github.com/bridgedotnet/Bridge/pull/1407#issuecomment-222538017)

LeonidVeriga:
The change is going to be postponed.

PR #1427, opened 2016-05-31
Bridge Builder: Default to a single csproj if found

### LINKED ISSUE #1132: Commandline builder: search for project file in current directory by default
Currently `Bridge.Builder.exe` requires an assembly or project file to be specified in the command line arguments in order to build the project.

Both `xbuild` and `msbuild`, if run from a project's root (or solution, in their case!) with no parameters defaults to building the project in the current configuration -- if and only if there's only one .sln or .csproj file in the directory.

If there are two or more eligible files (a .sln and a .csproj, or two .sln or two .csproj), it does not do anything with the message:
```
MSBUILD: error MSBUILD0005: Please specify the project or solution file to build, as more than one solution or project file was found in the current directory
```

Bridge commandline builder could have the same behavior -- but supporting only .csproj files -- in such a manner that having an event with just `build/Bridge.Builder.exe` would suffice, as it will load the current project on the current configuration (it would have to support reading `*.csproj.user` files for current configuration, or maybe reading environment variables if VS makes them available on events/tasks calls).

Phrohdoh:
Closes #1132

- Change the signature of `GetBridgeOptionsFromCommandLine` to `(ref string[], ILogger)` so that we may output the modified args instead of whatever was passed in.
- Only call `ShowHelp` when `GetBridgeOptionsFromCommandLine` returns `null`
- `Bridge.Builder.exe` will now default to a `csproj` in the target folder (either with the `-f` flag, or the cwd) if a single `csproj` is found

I will need help with this:
> (it would have to support reading *.csproj.user files for current configuration, or maybe reading environment variables if VS makes them available on events/tasks calls).

### DIFF
--- a/Compiler/Builder/Program.cs
+++ b/Compiler/Builder/Program.cs
@@ -16,10 +16,11 @@ namespace Bridge.Builder
         {
             var logger = new Logger(null, false, LoggerLevel.Info, true, new ConsoleLoggerWriter(), new FileLoggerWriter());
 
-            var bridgeOptions = GetBridgeOptionsFromCommandLine(args, logger);
+            var bridgeOptions = GetBridgeOptionsFromCommandLine(ref args, logger);
 
             if (bridgeOptions == null)
             {
+                ShowHelp(logger);
                 return 1;
             }
 
@@ -129,14 +130,8 @@ namespace Bridge.Builder
             return false; // didn't bind anywhere
         }
 
-        private static BridgeOptions GetBridgeOptionsFromCommandLine(string[] args, ILogger logger)
+        private static BridgeOptions GetBridgeOptionsFromCommandLine(ref string[] args, ILogger logger)
         {
-            if (args.Length == 0)
-            {
-                ShowHelp(logger);
-                return null; // error: arguments not provided, so can't guess what to do
-            }
-
             var bridgeOptions = new BridgeOptions();
 
             bridgeOptions.Name = "Bridge.Builder.Console";
@@ -153,7 +148,6 @@ namespace Bridge.Builder
                         if (bridgeOptions.Lib != null)
                         {
                             logger.Error("Error: Project and assembly file specification is mutually exclusive.");
-                            ShowHelp(logger);
                             return null;
                         };
                         bridgeOptions.ProjectLocation = args[++i];
@@ -218,7 +212,6 @@ namespace Bridge.Builder
                         if (bridgeOptions.ProjectLocation != null)
                         {
                             logger.Error("Error: Project and assembly file specification is mutually exclusive.");
-                            ShowHelp(logger);
                             return null;
                         }
                         bridgeOptions.Lib = args[++i];
@@ -264,7 +257,6 @@ namespace Bridge.Builder
                         if (!BindCmdArgumentToOption(args[i], bridgeOptions, logger))
                         {
                             logger.Error("Invalid argument: " + args[i]);
-                            ShowHelp(logger);
                             return null;
                         }
                         break;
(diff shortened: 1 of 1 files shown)


Compiler/Builder/Program.cs
```diff
@@ -275,8 +267,25 @@ private static BridgeOptions GetBridgeOptionsFromCommandLine(string[] args, ILog
 
             if (bridgeOptions.ProjectLocation == null && bridgeOptions.Lib == null)
             {
+                var folder = bridgeOptions.Folder ?? Environment.CurrentDirectory;
+                var csproj = "";
```
Phrohdoh: Looking at this again there is no reason to have this variable.
I can just shove `csprojs[0]` into the array initializer.

Compiler/Builder/Program.cs
```diff
@@ -291,4 +300,4 @@ private static BridgeOptions GetBridgeOptionsFromCommandLine(string[] args, ILog
             return bridgeOptions;
         }
     }
-}
```
Phrohdoh: Shall I undo this?

Compiler/Builder/Program.cs
```diff
@@ -275,8 +267,25 @@ private static BridgeOptions GetBridgeOptionsFromCommandLine(string[] args, ILog
...
+                if (csprojs.Length == 1)
+                {
+                    csproj = csprojs[0];
+                    args = new[] { "-f", folder, csproj };
+                    return GetBridgeOptionsFromCommandLine(ref args, logger);
+                }
+
+                if (csproj.Length > 1)
+                {
+                    logger.Warn("Could not default to a csproj because multiple were found.");
```
Phrohdoh: Would you like to me to list all of the files found?

Compiler/Builder/Program.cs
```diff
@@ -275,8 +267,25 @@ private static BridgeOptions GetBridgeOptionsFromCommandLine(string[] args, ILog
...
             if (bridgeOptions.ProjectLocation == null && bridgeOptions.Lib == null)
             {
+                var folder = bridgeOptions.Folder ?? Environment.CurrentDirectory;
+                var csproj = "";
+
+                var csprojs = Directory.GetFiles(folder, "*.csproj", SearchOption.TopDirectoryOnly);
+                if (csprojs.Length == 1)
+                {
+                    csproj = csprojs[0];
+                    args = new[] { "-f", folder, csproj };
```
LeonidVeriga: Can you please check if it should set the `-f` option? I would suggest directly setting `ProjectLocation`  to `sprojs[0]` so it avoids possible recursive calls of `GetBridgeOptionsFromCommandLine`.

Compiler/Builder/Program.cs
LeonidVeriga: Yes, it is a good idea to show as a comma-separated list.
As well as logging the project filename `csprojs[0]` in `if (csprojs.Length == 1)` to show that it was found auromatically.

Compiler/Builder/Program.cs
LeonidVeriga: You mean removing New Line? It would be nice.

LeonidVeriga:
Hi @Phrohdoh!

Thank you for this pull request. Welcome to Bridge :)

Before a review of the pull request we would need the CLA [form](http://goo.gl/forms/CvDFLA3EPO) accepted. You can find more information in [CONTRIBUTING.md](https://github.com/bridgedotnet/Bridge/blob/master/.github/CONTRIBUTING.md). 

The full CLA is available [here](http://bridge.net/cla/).

When you get a chance, please review the CLA and complete the [CLA Acceptance Form](http://goo.gl/forms/CvDFLA3EPO). If you have any questions, please just let us know.

Phrohdoh:
I have signed the CLA and will make the noted changes shortly.

Phrohdoh:
@LeonidVeriga This has been updated.

LeonidVeriga:
Thanks @Phrohdoh, we will review the pull request and try to get this into the Bridge 1.14 release.

LeonidVeriga:
@Phrohdoh, thank you for this nice pull request. It is merged and we will include this into 1.14 release.

Phrohdoh:
Awesome, thank you.

PR #1433, opened 2016-06-02
Fixes issue #1438 by adding support for JSON.ParseAsArray<SomeType>(jsonText)

### LINKED ISSUE #1432: JSON.Parse<SomeType[]>(jsonText) returns array of plain JS objects instead of Bridge class
### Expected

JSON.Parse<SomeType[]>(jsonText) should return array of SomeType instances.

```csharp
Deserialized with JSON.Parse<Foo[]>():
    I'm Demo.Foo and my value is 100
```

### Actual

It returns array of plain JS objects. Its properties are copied but without 'Bridge.merge()'-ing it with proper SomeType class instance.

```csharp
Deserialized with JSON.Parse<Foo[]>():
    result[0].someMethod is not a function
```

### Steps To Reproduce

http://deck.net/253736f96c7ddacdbe0c95fcf1383196

``` csharp
public class Foo
{
    public int Value { get; set; }

    public string SomeMethod()
    {
        return "I'm " + this.GetType().FullName + " and my value is " + this.Value;
    }
}

public class App
{
    public static void Main()
    {
        var serialized = JSON.Stringify(new[] { new Foo() { Value = 100 } });

        Console.WriteLine("Deserialized with JSON.Parse<Foo[]>():");
        var result = JSON.Parse<Foo[]>(serialized);
        TestFooInstance(result);
    }

    public static void TestFooInstance(Foo[] result)
    {
        try
        {
            Console.WriteLine("\t" + result[0].SomeMethod());
        }
        catch (Exception ex)
        {
            Console.WriteLine("\t" + ex);
        }
    }
}
```

To work around the issue, use `JSON.Parse<SomeType[]>(jsonText)` (see #1438) that returns array of  Bridge class instances.

### LINKED ISSUE #1438: Implement JSON.ParseAsArray<SomeType>(jsonText) to return array of Bridge classes
The issue created to implement a new `JSON.ParseAsArray` method to work around issue #1432 (`JSON.Parse<SomeType[]>(jsonText) returns array of plain JS objects instead of Bridge class`)

The implementation provided thanks to @d-p-y in PR #1433 

### Expected

`ParseAsArray` produces an array of `Foo` instances. The sample above outputs:
```
Deserialized with JSON.ParseAsArray<Foo>():
    I'm Demo.Foo and my value is 100
```

### Steps To Reproduce
Example([Live](http://live.bridge.net/#9eb022bd4e7cca9bf9a65e8c9315525a)):
```csharp
public class Foo
{
    public int Value { get; set; }

    public string SomeMethod()
    {
        return "I'm " + this.GetType().FullName + " and my value is " + this.Value;
    }
}

public class App
{
    public static void Main()
    {
        var serialized = JSON.Stringify(new[] { new Foo() { Value = 100 } });

        Console.WriteLine("Deserialized with JSON.Parse<Foo[]>():");
        var result = JSON.Parse<Foo[]>(serialized);
        TestFooInstance(result);

        Console.WriteLine("Deserialized with JSON.ParseAsArray<Foo>():");
        result = JSON.ParseAsArray<Foo>(serialized);
        TestFooInstance(result);
    }

    public static void TestFooInstance(Foo[] result)
    {
        try
        {
            Console.WriteLine("\t" + result[0].SomeMethod());
        }
        catch (Exception ex)
        {
            Console.WriteLine("\t" + ex);
        }
    }
}
```

d-p-y:
Fixes #1438 by supporting this particular deserialization scenario. I wanted to support IEnumerable<T> but for such general scenario deserialization of collections would require support for Reflection (guess we need to wait for Bridge v2 for that).

- Bridge.merge now has optional parameter(constructor of item prototype)
- added JSON.ParseAsArray<Item>(jsonText) that is an entry point for "deserialization of array of custom classes"
- added Bridge.Compiler.TranslatorTest to verify correctness

I considered modifying JSON.Parse<T>(jsonText) but it would require unsafe casts that's why it is implemented using another method.
As with my former bugfixes - Bridge.Testing is not available so I cannot guarantee it doesn't break anything.


LeonidVeriga:
Hi @d-p-y, thank you for this pull request, we will review the changes this week.

d-p-y:
Hi Leonid, I found out that my changes broke dictionary enumeration. It is fixed now

Html5/JSON.cs
```diff
@@ -20,12 +21,21 @@ public static class JSON
...
         /// <returns>The Object corresponding to the given JSON text.</returns>
         [Template("Bridge.merge(new {T}(), JSON.parse({text}))")]
         public static extern T Parse<T>(string text);
+        
+        /// <summary>
+        /// Arrays of custom classes compatible version of <see cref="Bridge.Html5.JSON.Parse&lt;T&gt;(System.String)"/>
+        /// </summary>
+        /// <param name="text">The string to parse as JSON. See the JSON object for a description of JSON syntax.</param>
+        /// <returns>The Object corresponding to the given JSON text.</returns>
+        [Template("Bridge.merge(new Array(), JSON.parse({text}), function(){return new {T}();})")]
```
kcudnik: > return new {T}();})")]

this assumes that T has default constructor which may not be true

d-p-y:
Kamil, You are absolutely right. Type constraint is added now.

LeonidVeriga:
Hi Dominik, we reviewed and merged this pull request. Thank you for contribution!

Please note, I created a separate issue #1438 to implement the new method `JSON.ParseAsArray` that is addressed by this pull request. Your original issue #1432 is open, I hope we will fix the original `JSON.Parse` method at some point.

Many thanks for the test you provided! I just moved it from Compiler tests into Client tests. It is more suitable for Client tests as it can be run and the actual runtime result can be compared to the expected one. We use Compiler tests as there is no difference in runtime results (or very hard to check) but the generated JavaScript matters.

d-p-y:
Leonid, 
That's fine for me! 
Indeed whole issue was about JSON.Parse<T>(string) being able to deserialize arrays. My code is just a working workaround(until we have support for Reflection).
Concerning Compiler vs Client tests - that's fine for me too. I used only available way to test for time being (till Bridge.Test becomes available).

Kind regards
Dominik

PR #1444, opened 2016-06-06
Weak map feature

### LINKED ISSUE #1443: Add support for HTML5 WeakMap
### Expected
It should be possible to use HTML5 WeakMap. It would be nice to have typesafe wrapper around it.

### Actual
Support for this feature is currently missing

### Steps To Reproduce
grepping for weakmap in sources of Bridge yields no result

d-p-y:
Fixes #1443.

- implements HTML5 WeakMap
- add clients tests (not sure if correct as not able to compile them without Bridge.Test)

I also propose to add such typesafe wrapper around WeakMap. I don't find any good place to add it (my understanding is that **Bridge.HTML5** may only contain delegates to native JS). 

```csharp
    /// <summary>
    /// Thin wrapper around <see cref="WeakMap"/>
    /// </summary>
    public class WeakDictionary<TKey,TValue> where TKey : class 
    {
        private WeakMap _impl;

        public WeakDictionary()
        {
             _impl = new WeakMap();
        }
        
        public WeakDictionary(IEnumerable<KeyValuePair<TKey,TValue>> items)
        {
             _impl = new WeakMap(
                 items.Select(x => new object[] {x.Key, x.Value}).ToArray()
             );
        }

        /// <summary>
        /// Clear all elements from Dictionary
        /// </summary>
        public void Clear()
        {
            //as WeakMap.clear() is not widely implemented it uses workaround
            _impl = new WeakMap();
        }
        public TValue this[TKey key]
        {
            get
            {
                return Get(key);
            }
            set
            {
                Set(key, value);
            }
        }

        public TValue Get(TKey key)
        {
            var result = _impl.Get(key);
            if (result == null)
            {
                // friendlier exception instead of default Bridge.InvalidCastException
                throw new KeyNotFoundException("No value for key present in WeakMap");
            }
            return (TValue)result;
        }

        public void Set(TKey key, TValue value)
        {
            _impl.Set(key, value);
        }

        public bool Delete(TKey key)
        {
            return _impl.Delete(key);
        }

        public bool ContainsKey(TKey key)
        {
            return _impl.Has(key);
        }
    }
```

### DIFF
--- a/Html5/WeakMap.cs
+++ b/Html5/WeakMap.cs
@@ -11,17 +11,13 @@
         /// <summary>
         ///  Default constructor
         /// </summary>
-        public WeakMap()
-        {
-        }
-
+        public extern WeakMap();
+        
         /// <summary>
         ///  Initialize WeakMap using items that are expected to be an array of the two-element-arrays. Each two-element-array is key and value respecively
         /// Beware that it is not widely implemented (2016-05)
         /// </summary>
-        public WeakMap(object[][] items)
-        {
-        }
+        public extern WeakMap(object[][] items);
 
         /// <summary>
         /// gets contained element OR returns null. 
--- a/Html5/WeakMap.cs
+++ b/Html5/WeakMap.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Bridge.Html5
+{
+    /// <summary>
+    /// HTML5 WeakMap. Keys must be object (primitives are not allowed). Values can be any type (inc. primitives)
+    /// <a href="https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/WeakMap">MDN</a>
+    /// </summary>
+    [External]
+    [Name("WeakMap")]
+    public class WeakMap
+    {
+        /// <summary>
+        ///  Default constructor
+        /// </summary>
+        public WeakMap()
+        {
+        }
+
+        /// <summary>
+        ///  Initialize WeakMap using items that are expected to be an array of the two-element-arrays. Each two-element-array is key and value respecively
+        /// Beware that it is not widely implemented (2016-05)
+        /// </summary>
+        public WeakMap(object[][] items)
+        {
+        }
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


Html5/WeakMap.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+        }
+
+        /// <summary>
+        ///  Initialize WeakMap using items that are expected to be an array of the two-element-arrays. Each two-element-array is key and value respecively
+        /// Beware that it is not widely implemented (2016-05)
+        /// </summary>
+        public WeakMap(object[][] items)
+        {
+        }
```
kcudnik: constructors also can be marked as extern

Html5/WeakMap.cs
```diff
@@ -0,0 +1,49 @@
...
+        public virtual extern void Set<T>(T key, object value) where T : class;
+
+        /// <summary>
+        /// deletes element. Returns true if element was contained in the map. Returns false when it wasn't present in the map
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns></returns>
+        public virtual extern bool Delete<T>(T key) where T : class;
+
+        public virtual extern bool Has<T>(T key)  where T : class;
```
kcudnik: Also add summary documentation headers to be consistent ?

d-p-y:
Kamil, I've changed code according to your remarks

d-p-y:
Hi, 
Question: are those pushed tests Tests/Batch1/Collections/Native/WeakMapTests.cs
 useful or I should not bother in future? I ask because I have yet another submission ready and I wonder if I should work on this or not.
Thanks
Dominik

LeonidVeriga:
Hi @d-p-y!

Sorry for that delay in responding.
Yes, the tests you commited are very useful.

LeonidVeriga:
Hi @d-p-y,

Merged your Pull Request by https://github.com/bridgedotnet/Bridge/commit/ad8bc27eb00db640b89426c2e1bd3fc6b1d0167b.

Thank you for your time and contribution.

LeonidVeriga:
@d-p-y,
Regarding your suggestion on `WeakDictionary`, we think that is a good idea and you could create your own NuGet package with name like `<your brandname>.<your utils/extention name>`

Issue #1453, opened 2016-06-08
String.Remove should throw Exception is index out of range


geoffreymcgill:
The following `String.Remove()` samples should throw JavaScript Exceptions

```c#
var val = "Hello";

Console.WriteLine(val.Remove(5));
Console.WriteLine(val.Remove(0, 6));
Console.WriteLine(val.Remove(6, 0));

val = null;
Console.WriteLine(val.Remove(1, 1));
```

michaelcheers:
Related to #72.

LeonidVeriga:
Related #554

Issue #1447, opened 2016-06-06
String.Remove(0, 0) returns empty string


Suchiman:
### Expected
`"Hello World".Remove(0, 0)` returning `"Hello World"`

### Actual
`"Hello World".Remove(0, 0)` returning `""`

### Steps To Reproduce
http://live.bridge.net/#83a008e47570da901f14cd63f4c5b7ae
```csharp
public class App
{
    [Ready]
    public static void Main()
    {
        Window.Alert("Hello World".Remove(0, 0));
    }
}
```

geoffreymcgill:
Thanks for reporting this issue. There must be a logic error in the following js function:

https://github.com/bridgedotnet/Bridge/blob/master/Bridge/Resources/String.js#L384

TriEdgeAI:
Shouldn't be `!count`, because that would result in `true` when `count` is 0.

geoffreymcgill:
Looks like there are several other issues within that function. I'll try and fix and sneak into the 1.14 release.

geoffreymcgill:
This issue has been fixed and will be included in the 1.14 release.

Also related to #1453.

Issue #1106, opened 2016-03-25
Replace Bridge.hasValue by != null construction


vladsch:
http://forums.bridge.net/forum/general/feature-requests/1910

DaniilVeriga:
It looks like there is one more related question:
http://forums.bridge.net/forum/community/help/1926

PR #1454, opened 2016-06-08
[#1447] [#1453] String.Remove(0, 0) returns empty string

geoffreymcgill:
Adderesses #1447, #1453


Bridge/Resources/String.js
```diff
@@ -382,7 +382,7 @@
         },
 
         remove: function (s, index, count) {
-            if (!count || ((index + count) > this.length)) {
+            if (!Bridge.hasValue(count) || ((index + count) > s.length)) {
```
michaelcheers: @geoffreymcgill, why not use == null instead of Bridge.hasValue #1106.

PR #1463, opened 2016-06-09
Fixed #1344 (try 2, tested)

### LINKED ISSUE #1344: Fields cannot be named __proto__
### Expected
```
True
```
### Actual
```javascript
false
```
### Steps To Reproduce
[Live Bridge](http://live.bridge.net/#344ddb787193fdf18fa2ac95f1d97afb)
```csharp
/// <summary>
/// Creates a simple class
/// </summary>
public class App
{
    public string __proto__;
    
    public App ()
    {
        __proto__ = "hello world";
        Console.WriteLine(__proto__ == "hello world");
    }
    
    [Ready]
    public static void Main()
    {
        App app = new App();
    }
}
```

michaelcheers:
Fixed #1344.

**PS.** I have no idea how to test out whether the code I wrote works or not.
**Update.** It works (tested with Bridge Live) :smile:

### DIFF
--- a/Compiler/Contract/Constants/JS.cs
+++ b/Compiler/Contract/Constants/JS.cs
@@ -168,8 +168,8 @@
 
         public class Reserved
         {
-            public static readonly List<string> StaticNames = new List<string> { "Name", "Arguments", "Caller", "Length", "Prototype" };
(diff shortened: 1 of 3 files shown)


LeonidVeriga:
`__proto__`  is a "proper" field and may be considered as a "proprietary".
We are thinking of creating and separate extension class or adding properties into `Object` class.

An issue #2082 created.

PR #1496, opened 2016-06-24
[#1485] Performance problems with constructors

### LINKED ISSUE #1485: Performance problems with constructors
We are developing a project which translates existing Unity games, written in C# to JavaScript and 2D canvas. Previously we were using Saltarelle, but we decided to move on to Bridge.NET, because of the acquisition and all. However once we started testing we saw very low framerates in the resulting builds and we decided to go back to Saltarelle just to see what's going to happen. That increased the framerate 5 to 10 times in all games. After some analysis we discovered that Saltarelle generates much simpler constructors than Bridge.NET and that might be one of the main issues for the way lower performance with Bridge.NET.

Take for example the following test case:

```csharp
public class App
{
    public struct Vector3
    {
        public float x;
        public float y;
        public float z;
        
        public Vector3(float x, float y, float z)
        {
            this.x = x;
            this.y = y;
            this.z = z;
        }
        
        public static Vector3 operator+(Vector3 a, Vector3 b)
        {
            return new Vector3(a.x + b.x, a.y + b.y, a.z + b.z);
        }
    }
    
    public static void Main()
    {
        DateTime start = DateTime.Now;
        
        Vector3 sum = new Vector3(0.0f, 0.0f, 0.0f);
           
        for(int i = 0; i < 1000 * 1000; i++)
        {
            sum += new Vector3(1.0f, 2.0f, 3.0f);
        }
        
        DateTime end = DateTime.Now;
        
        Console.Write((end - start).TotalMilliseconds);
    }
}
```

The generated constructor by Bridge.NET would be:
```javascript
function Class() {
    if (!(this instanceof Class)) {
        var args = Array.prototype.slice.call(arguments, 0),
            object = Object.create(Class.prototype),
            result = Class.apply(object, args);

        return typeof result === "object" ? result : object;
    }

    // All construction is actually done in the init method
    if (!initializing) {
        if (this.$staticInit) {
            this.$staticInit();
        }

        if (this.$initMembers) {
            this.$initMembers.apply(this, arguments);
        }

        this.$$initCtor.apply(this, arguments);
    }
}
```

which involves multiple branches, string operations and so on, while Saltarelle simply generates:

```javascript
$App$Vector3.$ctor1 = function(x, y, z) {
	this.x = 0;
	this.y = 0;
	this.z = 0;
	this.x = x;
	this.y = y;
	this.z = z;
};
```

And here are the average benchmark results for the above snippet on Chrome on my Acer Aspire V3-772G laptop:
Bridge.NET: ~55,000 ms
Saltarelle: ~26 ms

Are you guys perhaps going to consider changing the way objects are constructed with Bridge.NET? I apologize if submitting this issue sounds like a complaint, but it is a major roadblock for moving from Saltarelle to Bridge.NET.

vladsch:
Fixes #1485


LeonidVeriga:
Instance creation time reduced (a sample creating 10^6 struct instances [Live Bridge](http://live.bridge.net/#4d88e43a28f95fdb1cedb7f22f10ce7d)) - apprx 80ms now vs 20s (250 times faster); marking struct with `[Immutable]` gives apprx 40ms vs 19s (500 times faster).
 
Concomitant changes:
 - Class constructor naming `constructor$1` -> `$constructor1`
 - Constructor call syntax `new Demo.App("constructor", ...)` -> `new Demo.App.$constructor(...)`
 - Each constructor has the following code in the beginning `this.$initialize();`

PR #1505, opened 2016-06-29
[#1504] Support of constructor body wrpapping by injectors from plugin

### LINKED ISSUE #1504: Support of constructor body wrpapping by injectors from plugin

vladsch:
Fixes #1504.


LeonidVeriga:
A Plugin can wrap a constructor being emitted (in `ConstructorBlock` class). In order to do that, the Plugin's  `IPlugin.GetConstructorInjectors` method should return a string prefixed with `$ctorWrapper:` and the followed a JavaScript code that wraps up type constructors.

PR #1655, opened 2016-08-11
[#1379] Double (s) and Float (s) have String constants for Nan, Infin…

### LINKED ISSUE #1379: Double (s) and Float (s) have String constants for Nan, Infinity, Negative Infinity,
### Expected

System.Double  (or maybe Bridge.Double considering #1293) (3 times)
System.Single (or maybe Bridge.Double considering #1293) (3 times)

### Actual

String (6 times)

### Steps To Reproduce
[Live Bridge](http://live.bridge.net/#f82a3f28b658e025d37e2b1574c1dcbb)
[Dot Net Fiddle](https://dotnetfiddle.net/Z8lVf2)
```csharp
/// <summary>
/// Creates a simple class
/// </summary>
public class App
{
    [Ready]
    public static void Main()
    {
        object value = null;
        for (int time = 0; time < 6; time++)
        {
            switch (time)
            {
                case 0:
                {
                    value = 0.0 / 0.0;
                    break;
                }
                case 1:
                {
                    value = 1.0 / 0.0;
                    break;
                }
                case 2:
                {
                    value = -1.0 / 0.0;
                    break;
                }
                case 3:
                {
                    value = 0.0f / 0.0f;
                    break;
                }
                case 4:
                {
                    value = 1.0f / 0.0f;
                    break;
                }
                case 5:
                {
                    value = -1.0f / 0.0f;
                    break;
                }
            }
            Console.WriteLine(value.GetType().FullName);
        }
    }
}
```

vladsch:
Fixes #1379.


LeonidVeriga:
In the current implementation, when getting type for consts:
```cs
0.0 / 0.0;
1.0 / 0.0;
1.0 / 0.0;
0.0f / 0.0f;
1.0f / 0.0f;
1.0f / 0.0f;
```
They are considered to be `System.Double`

Issue #1538, opened 2016-07-23
Out parameters within indexer code fails


danieldegtyarev:
### Expected

To work without errors

### Actual

Uncaught TypeError: Cannot set property 'v' of undefined in

```csharp
tryGetValue: function (key, value) {
      var entry = this.findEntry(key);

       value.v = entry ? entry.value : Bridge.getDefaultValue(TValue);

      return !!entry;
}
```

### Steps To Reproduce

http://live.bridge.net/#34c0f1bfd27bceb61f001b5b4c8b1b96

```csharp
internal class Map<TValue> : Dictionary<string, TValue>
{
    readonly Func<string, TValue> defaultValueFactory;

    public Map(Func<string, TValue> defaultValueFactory = null)
    {
        this.defaultValueFactory = defaultValueFactory;
    }

    public new TValue this[string key]
    {
        get
        {
            TValue result;
            if (!TryGetValue(key, out result))
            {
                result = defaultValueFactory == null ? default(TValue) : defaultValueFactory(key);
                Add(key, result);
                return result;
            }
            return result;
        }
    }
}

/// <summary>
/// Creates a simple class
/// </summary>
public class App
{
    public static void Main()
    {
        Map<int> foo = new Map<int>(x => 1);
        var bar = foo["something"];
    }
}
```

LeonidVeriga:
Hi @danieldegtyarev,

That is a very nice catch. Thank you for reporting this.

We narrowed down the problem - related to handling calls with `out` parameters within indexers.
Sample ([Bridge Live](http://live.bridge.net/#be809d4e85d966b711e26a75d6514250)):
```
internal class ClassA<T>
{
    private static void GetSomething(out T result)
    {
        result = default(T);
    }

    public new T this[string key]
    {
        get
        {
            // result variable should be an object {}
            T result;
            GetSomething(out result);
            
            // return should be result.v
            return result;
        }
    }
}

public class App
{
    public static void Main()
    {
        var c = new ClassA<int>();
        
        var r = c[""];
    }
}
```

vladsch:
Fixed in https://github.com/bridgedotnet/Bridge/pull/1679

PR #1679, opened 2016-08-16
[#1519] Number variables with the same name but in different scope ha…

### LINKED ISSUE #1519: Number variables with the same name but in different scope have incorrect type
Reorted on the forums http://forums.bridge.net/forum/bridge-net-pro/bugs/2481

### Expected
```
Inside if statement scope: 1
Value: 1
Value: 2
```

### Actual
```
Inside if statement scope: 1
Value: undefined
Value: undefined
```

### Steps To Reproduce
Sample:
[Bridge Live](http://live.bridge.net/#80132f207a5f35fe35a5c1dcfae3fa64)
[.Net Fiddle](https://dotnetfiddle.net/Y1bIhc)
```csharp
public class App
{
    public static void Main()
    {
        bool boolean = true;
        Dictionary<int, int> dic = new Dictionary<int, int>();
        dic.Add(1, 1);
        dic.Add(2, 2);
        
        if (boolean) {
            int sameVal;
            if (dic.TryGetValue(1, out sameVal)) {
                Console.WriteLine("Inside if statement scope: " + sameVal);
            }
        }
        
        foreach(int sameVal in dic.Values) {
            Console.WriteLine("Value: " + sameVal);
        }
    }
}
```

vladsch:
Fixes #1519


vladsch:
Added fix for #1526 (related functionality)

vladsch:
The pull request fixes https://github.com/bridgedotnet/Bridge/issues/1538 also

PR #1688, opened 2016-08-18
[#1533] String's are by default null not ""

### LINKED ISSUE #1533: String's are by default null not ""
### Expected

b

### Actual

nullb

### Steps To Reproduce
[Live Bridge](http://live.bridge.net/#0025131317bb2e5d573bd6ac0aa62430)
[Dot Net Fiddle](https://dotnetfiddle.net/PNV6bt)
```csharp
/// <summary>
/// Creates a simple class
/// </summary>
public class App
{
    public static void Main()
    {
		Console.WriteLine(@string += 'b');
    }
	
	static string @string;
}
```

vladsch:
Fixes #1533 
String + operator is implemented via string array join method now to convert null to empty string


michaelcheers:
@vladsch, with the more bugs we fix the harder the code is to read.

Issue #1310, opened 2016-05-07
Change System.Console class


michaelcheers:
I was thinking that it would be cool to change the System.Console class so instead of printing to the javascript console it would print to a javascript 
```
<p>
```
 by default but it could print to any TextWriter like in C#. I am thinking we keep Bridge.Console the same but change System.Console to a TextWriter with a default of a 
```
<p>
```
.
There are many methods that can't be implemented in the Console.
**For example:**
```csharp
Console.Write // (can only be implemented with Console.Log)
Console.SetOut
Console.SetIn
Console.SetError
Console.CancelKeyPress
Console.SetWindowSize
Console.SetWindowPosition
Console.SetCursorPosition
Console.SetBufferSize
Console.ResetColor
Console.ReadLine // (only with prompt which isn't actually the console)
Console.ReadKey
Console.Read
Console.OpenStandardError
Console.OpenStandardInput
Console.OpenStandardOutput
Console.MoveBufferArea
Console.Clear
```

geoffreymcgill:
Hi @michaelcheers,

Yes, I agree with everything you have noted here. The functionality of **System.Console** has been annoying me lately and we are going to change.

I was thinking almost the exact same thing, that **Bridge.Console** would continue to log to the browser Console, but **System.Console** would log to some kind of new visible output window.

We're going to mock up some ideas and try to implement for the Bridge 15 release.

geoffreymcgill:
We discussed this feature request today and have a plan to implement. Should be available for Bridge 15.

geoffreymcgill:
This issue is going to be closed soon, although only partially implemented with the new Bridge.Console functionality. 

We will continue to expand on the functionality, and track in issue #1763.

geoffreymcgill:
Bridge Console has been merged into the *master* branch.

PR #1762, opened 2016-09-06
Bridge.Console 

geoffreymcgill:
Implements a custom Bridge Console which renders log messages directly in the browser document without having to open the browser developer tools. 

Related to #1310.


Bridge/Bridge/Console/Console.cs
```diff
@@ -0,0 +1,473 @@
...
+        {
+            LogBase(value);
+        }
+
+        public static void Log(string value)
+        {
+            LogBase(value);
+        }
+
+        public bool Hidden = true;
```
kcudnik: Hidden have default setter, if you set it to any value there will be no action taken, should that be property with private set ?

geoffreymcgill: Hi @kcudnik - Yes, `Hidden` should be `private`. I have updated the branch. 

Thanks for pointing out the issue.

Issue #1633, opened 2016-08-07
IFormattable - missing ToString (Format method instead)


LeonidVeriga:
[MSDN IFormattable Interface](https://msdn.microsoft.com/en-us/library/system.iformattable(v=vs.110).aspx)

### Expected
[MSDN IFormattable.ToString Method (String, IFormatProvider)](https://msdn.microsoft.com/en-us/library/system.iformattable.tostring(v=vs.110).aspx)
```csharp
string ToString(
	string format,
	IFormatProvider formatProvider
)
```

### Actual
```csharp
public string Format(
	string format,
	IFormatProvider formatProvider
)
```

PS Should be addressed along with #1565

calcock:
Hello,
Here are some  interesting facts that may be useful for you, I think this information could be  helpful, read it  here  <http://support.noisecircuit.net/lnskh/4>

Be well, calcock

PR #1782, opened 2016-09-10
[1565] IFormattable - Interface implementation does not invoke right …

### LINKED ISSUE #1565: IFormattable - Interface implementation does not invoke right method
### Steps To Reproduce

```csharp
public class App
{
    public static void Main()
    {
        Assert.AreEqual(
            ((IFormattable)new MyFormattable()).Format("real", new MyFormatProvider()),
            "real success, The provider",
            "Interface call should succeed");
    }

   	private class MyFormattable : IFormattable
   	{
		public string ToString(string format, IFormatProvider provider)
		{
		    return format + " success, " + provider.GetFormat(typeof(object));
		}
   	}
	
	private class MyFormatProvider : IFormatProvider
   	{
		public object GetFormat(Type formatType)
		{
		    return "The provider";
		}
   	}
}
```

vladsch:
Fixes #1565


LeonidVeriga:
Fixes #1633

PR #1946, opened 2016-10-09
[1907] Modify Bridge.hasValue to accept multiple arguments

### LINKED ISSUE #1907: Modify Bridge.hasValue to accept multiple arguments
**Proposal** to add functionality to `Bridge.hasValue` which would allow passing 0 to many argument values. Each value would be checked. If all arguments have a value, then `true` is returned, otherwise `false` is returned.

This change would add a little more logic to the `hasValue` functionality, but would reduce the **.js** size in many places throughout **bridge.js**.

### Expected

```js
Bridge.hasValue(a, b) // all true, or return false
```

### Actual

```js
Bridge.hasValue(a) && Bridge.hasValue(b)
```

No breaking changes would occur. This change would be enhance the functionality of the existing `hasValue` function.

LeonidVeriga:
Fixes #1907.

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -457,6 +457,26 @@
             return Bridge.Reflection.getTypeFullName(obj);
         },
 
+        hasValue: function (obj) {
+            return obj != null;
+        },
+
+        hasValue$1: function () {
+            if (arguments.length === 0) {
+                return false;
+            }
+
+            var i = 0;
+
+            for (i; i < arguments.length; i++) {
+                if (arguments[i] == null) {
+                    return false;
+                }
+            }
+
+            return true;
+        },
+
         is: function (obj, type, ignoreFn, allowNull) {
             if (type && type.prototype && type.prototype.$literal && Bridge.isPlainObject(obj)) {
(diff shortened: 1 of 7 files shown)


LeonidVeriga:
@geoffreymcgill,

**1)** Optimizes existing `Bridge.hasValue` (faster **105 ms** against **189 ms** on your perf test [Deck](http://deck.net/aac7c465582442a75f024b95dcd30cc9/left))
Current:
```js
hasValue: function (obj) {
    return (obj !== null) && (obj !== undefined);
},
```
Revised:
```js
hasValue: function (obj) {
    return obj != null;
},
```
**2)** Adds an overload `Bridge.hasValue$1` to work with function `arguments`, a trade-off to keep exisint fast `hasValue` and save some bytes for **bridge.js** in case of multiple arguments:
```js
hasValue$1: function () {
    if (arguments.length === 0) {
        return false;
    }

    var i = 0;

    for (i; i < arguments.length; i++) {
        if (arguments[i] == null) {
            return false;
        }
    }

    return true;
},
```

Saves 259 byte savings in the **bridge.js** file size: 259 bytes = 883'226 - 882'967

PR #1975, opened 2016-10-20
[1670] JsDictionary - support a helper for JavaScript objects to use …

### LINKED ISSUE #1670: JsDictionary - support a helper for JavaScript objects to use as dictionary
The functionality of the **JsDictionary** class is covered by the **Object** class in Bridge. 

We will not be implementing **JsDictionary**, unless someone comes up with a good reason. The only good reason we can think of is to avoid a Breaking Change when upgrading from Saltarelle to Bridge.

```C#
/// <summary>
/// The JsDictionary data type which is mapped to the Object type in Javascript.
/// </summary>
[External]
[Namespace(false)]
[Name("Object")]
public sealed class JsDictionary : IEnumerable
{
    [Template("{{}}")]
    public extern JsDictionary();

    public extern JsDictionary(params object[] nameValuePairs);

    public extern int Count
    {
        [Template("{$System.Script}.getKeyCount({this})")]
        get;
    }

    public new extern ICollection<string> Keys
    {
        [Template("{$System.Object}.keys({this})")]
        get;
    }

    public extern object this[string key]
    {
        get;
        set;
    }

    [Template("{$System.Script}.clearKeys({this})")]
    public extern void Clear();

    [Template("{$System.Script}.keyExists({this}, {key})")]
    public extern bool ContainsKey(string key);

    [NonScriptable]
    public static extern JsDictionary GetDictionary(object o);

    [Template("new {$System.ObjectEnumerator`2}({this})")]
    public extern Enumerator<string, object> GetEnumerator();

    extern IEnumerator IEnumerable.GetEnumerator();

    [Template("delete {this}[{key}]")]
    public extern void Remove(string key);

    [Template("{this}[{key}] = {value}")]
    public extern void Add(string key, object value);

    [NonScriptable]
    public static extern implicit operator JsDictionary<string, object>(JsDictionary value);

    [NonScriptable]
    public static extern implicit operator JsDictionary(JsDictionary<string, object> value);
}
```

vladsch:
Fixes #1670

Stack and Queue implementations are added also to Bridge.Collections.Generic (simple implementation without additional client side code)

### DIFF
--- a/Bridge/Bridge.csproj
+++ b/Bridge/Bridge.csproj
@@ -83,6 +83,10 @@
     <Compile Include="Attributes\ExternalAttribute.cs" />
     <Compile Include="Attributes\FileNameAttribute.cs" />
     <Compile Include="Attributes\ScriptAttribute.cs" />
+    <Compile Include="Bridge\Collections\Generic\JsDictionary.cs" />
+    <Compile Include="Bridge\Collections\Generic\Queue.cs" />
+    <Compile Include="Bridge\Collections\Generic\Stack.cs" />
+    <Compile Include="Bridge\Collections\JsDictionary.cs" />
     <Compile Include="Bridge\Console\Console.cs" />
     <Compile Include="Bridge\Function.cs" />
     <Compile Include="Bridge\Linq\Enumerable.cs" />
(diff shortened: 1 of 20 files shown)


LeonidVeriga:
The PR is on-hold until the feature is requested by community. The branch is supposed to be kept.

It contains a fix for #1996 that is extracted into separate PR #1997 and merged.

PR #1987, opened 2016-10-24
[1081] Add millisecond format support to TimeSpan

### LINKED ISSUE #1081: TimeSpan millisecond format unsupported
### Expected
```
00:00.000
```

"f" is unsupported as a formatter. Notably, this is handled for DateTime, but not TimeSpan. Also, escaped characters are not emitted in .NET whereas are here.

[MSDN The"f" Custom Format Specifier](https://msdn.microsoft.com/en-us/library/ee372287(v=vs.110).aspx#fSpecifier)
### Actual
```
00\:00\.fff
```
### Steps To Reproduce

[Deck](http://deck.net/a7ad73a38ccb4b897fa246b3698228f8)

``` csharp
public class App
{
    [Ready]
    public static void Main()
    {
        var s = new TimeSpan(0).ToString(@"mm\:ss\.fff");

        Console.WriteLine(s);
    }
}
```

haxard:
Fixes #1081 .

- New recognition of the "f", "ff", "fff" format syntax for TimeSpan

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -1659,8 +1659,8 @@
             return braces.substr(0, (braces.length + (remove ? 0 : 1)) / 2);
         },
 
-        alignString: function (str, alignment, pad, dir) {
-            if (!alignment) {
+        alignString: function (str, alignment, pad, dir, cut) {
+            if (!str || !alignment) {
                 return str;
             }
 
@@ -1677,6 +1677,11 @@
             }
 
             alignment = Math.abs(alignment);
+
+            if (cut && (str.length > alignment))
+            {
+                str = str.substring(0, alignment);
+            }
 
             if (alignment + 1 >= str.length) {
                 switch (dir) {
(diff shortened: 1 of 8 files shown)


haxard:
Sorry about that. Reverted and checked in one with spaces.

Not sure why GitHub is saying so many changes took place. VS doesn't look like that

Bridge/Resources/.generated/bridge.js
```diff
@@ -8007,254 +8007,259 @@
 
     // @source TimeSpan.js
 
-    Bridge.define("System.TimeSpan", {
-        inherits: [System.IComparable],
+Bridge.define("System.TimeSpan", {
```
LeonidVeriga: > Not sure why GitHub is saying so many changes took place. VS doesn't look like that

@haxard, can you please add the original four spaces (i.e. shifting the text to the right on four spaces).

We recently changed our committing code policy not to reference issues in commit descriptions.
Instead of using `[#nnnn]` or `bridgedotnet/Bridge#nnnn` just use `[nnnn]` or `bridgedotnet/Bridgennnn`. This keeps all commits history under corresponding pull requests and keeps issue history plain and clean.

haxard:
Should be good now

LeonidVeriga:
Finally, merged with the following additional commits:
- 1b3f652 Fix TimeSpan.js format regexp to be greedy and search for seven `f`
- 67eeede Fix TimeSpan f format being not escaped, add client test for escaping
- Added client tests

The squash commit containg overall changes - 0c8640b

PR #2004, opened 2016-10-31
[2003] lambda expressions referencing 'this' are not compiled correctly

### LINKED ISSUE #2003: lambda expressions referencing 'this' are not compiled correctly
### Expected

Lambda expressions containing calls to natively implemented member should be compiled correctly. In other words: if 'this' is used within anonymous method then 'method.bind(this)' is generated.

### Actual

Actually such lambda expressions have no 'bind' issued. Because of that 'this' is null.

### Steps To Reproduce

[Deck](http://deck.net/0120918fe145c1b2e4613b3825903169)

```cs
public class Test
{
    [Template("{this}.setSomeProp({this}.getSomeProp() + 1)")]
    public extern object SomeInline();

    public int SomeProp
    {
        get; set;
    }

    public void Run()
    {

        Func<object> cb = () => { return SomeInline(); };
        SomeProp = 5;
        try
        {
            cb();
            Console.WriteLine("bug[1]?=" + (6 != SomeProp));
        }
        catch (Exception ex)
        {
            Console.WriteLine("bug[2]=" + ex);
        }
    }

    public static void Main()
    {
        new Test().Run();
    }
}
```

I'll send pull request in the moment.

d-p-y:
Fixes #2003

-Make lambda expressions increase Emitter's ThisRefCounter if 'this' is detected in generated template script
-Test added to avoid regression in future (but not sure if it works as Bridge.Testing is not available)

### DIFF
--- a/Compiler/Translator/Emitter/Blocks/IdentifierBlock.cs
+++ b/Compiler/Translator/Emitter/Blocks/IdentifierBlock.cs
@@ -134,6 +134,7 @@ namespace Bridge.Translator
 
             if (hasThis)
             {
+                Emitter.ThisRefCounter++;
                 this.Write("");
                 var oldBuilder = this.Emitter.Output;
                 this.Emitter.Output = new StringBuilder();
--- a/Compiler/Translator/Emitter/Blocks/InvocationBlock.cs
+++ b/Compiler/Translator/Emitter/Blocks/InvocationBlock.cs
@@ -203,6 +203,11 @@ namespace Bridge.Translator
                             inlineScript = inlineScript.Substring(6);
                         }
 
+                        if (!noThis)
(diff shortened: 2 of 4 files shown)


d-p-y:
ok, I've added requested change

LeonidVeriga:
Hi @d-p-y,

We are accepting this Pull Request. We appreciate you add the client tests.

The only thing we changed is adding the test into **Batch3** project.

We also changed a bit our commit policy not to reference the issue being fixed in commit messages (using **#**). It keeps the issue history clean (referencing just PR works perfect).

d-p-y:
Thanks for explaining logic behind those batches. Regarding changes - they are good for me.

PR #2014, opened 2016-11-02
Issue2013 fixes event subscription

### LINKED ISSUE #2013: Defect: subscription to event defined in generic interface via extension method silently fails to subscribe
### Expected

Subscribing to event should work.

### Actual
It silently fails. Using 'Steps to Reproduce' generated code should be 
``js
self.Ev = Bridge.fn.combine(self.Ev,$_.Demo.EvExtension.f1);
``

generated code is:
``js
self["Demo$IEvGen$1$" + Bridge.getTypeAlias(T) + "$Ev"] = Bridge.fn.combine(self["Demo$IEvGen$1$" + Bridge.getTypeAlias(T) + "$Ev"], $_.Demo.EvExtension.f1);
``

Pull request will be submitted in a while

### Steps To Reproduce

[Deck](http://deck.net/c4301542f49f4390a3a6fa4c03f141e4)

```cs

    public interface IEvGen<T>
    {
        event Action Ev;
    }
    
    public class EvGen<T> : IEvGen<T>
    {
        public event Action Ev;
        public bool HasListeners {get {return Ev != null;}}
    }
    
    public static class EvExtension
    {
        public static void AttachViaExtension<T>(this IEvGen<T> self)
        {
            self.Ev += () => {};
        }
    }

    public class Program { 
        
	[Ready]
        public static void Main()
        {
             var sut = new EvGen<int>();
			 
            IEvGen<int> sutIface = new EvGen<int>();
            sutIface.AttachViaExtension();
                        
            Console.WriteLine("bug? " + !sut.HasListeners);
        }
    }
```

d-p-y:
Fixes #2013

- change implementation to generate JS code as described in issue
- added Translator tests that hopefully work (no idea about it without Bridge.Test)

### DIFF
--- a/Compiler/Translator/Emitter/Blocks/AssignmentBlock.cs
+++ b/Compiler/Translator/Emitter/Blocks/AssignmentBlock.cs
@@ -291,24 +291,22 @@ namespace Bridge.Translator
 
                     if (leftMemberResolveResult != null)
                     {
-                        isEvent = leftMemberResolveResult.Member is DefaultResolvedEvent;
+                        isEvent = leftMemberResolveResult.Member is DefaultResolvedEvent ||
+                             leftMemberResolveResult.Member is SpecializedEvent;
                     }
+                    
+                    this.Emitter.IsAssignment = true;
+                    this.Emitter.AssignmentType = AssignmentOperatorType.Assign;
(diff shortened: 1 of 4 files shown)


LeonidVeriga:
Covered by https://github.com/bridgedotnet/Bridge/pull/2022

PR #2041, opened 2016-11-09
Fixed #2034: Window.OnError wrong arguments.

### LINKED ISSUE #2034: Window.OnError wrong arguments
`Window.OnError` is `Action<Event>`, but it should be `Action<string, string, int, int, Error>` according to:
https://developer.mozilla.org/en/docs/Web/API/GlobalEventHandlers/onerror

TriEdgeAI:
Fixed #2034: Window.OnError wrong arguments.

### DIFF
--- a/Html5/Elements/HTMLBodyElement.cs
+++ b/Html5/Elements/HTMLBodyElement.cs
@@ -62,7 +62,7 @@ namespace Bridge.Html5
         /// Exposes the window.onerror event handler to call when the document fails to load properly. Note: This handler is triggered when the event reaches the window, not the body element. Use addEventListener() to attach an event listener to the body element.
         /// </summary>
         [Name("onerror")]
-        public new Action<Event> OnError;
+        public new ErrorEventHandler OnError;
 
         /// <summary>
         /// Exposes the window.onfocus event handler to call when the window gains focus. Note: This handler is triggered when the event reaches the window, not the body element. Use addEventListener() to attach an event listener to the body element.
--- a/Html5/ErrorEventHandler.cs
+++ b/Html5/ErrorEventHandler.cs
@@ -2,5 +2,5 @@ using System;
 
 namespace Bridge.Html5
 {
-    public delegate bool ErrorEventHandler(string message, string url, int lineNumber, int columnNumber, Exception error);
+    public delegate bool ErrorEventHandler(string message, string url, int lineNumber, int columnNumber, Error error);
 }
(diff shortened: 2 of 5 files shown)


LeonidVeriga:
Hi @TriEdgeAI, thanks for the Pull Request.

Looks like it can be even improved:
- `WindowInstance.OnError Action<Event>` should be also changed;
- There is an existing `ErrorEventHandler` that can be used instead of `Action<string, string, int, int, Error>`
  And there is a bug with `ErrorEventHandler` itself (uses last parameter as `Exception` instead of `Error`) that your PR has discovered (:+1:).

TriEdgeAI:
Oh, yes. I was going to suggest using a delegate instead of Action<> in this case, because we will have names for the arguments. Definitely a better idea.

TriEdgeAI:
How about now?

TriEdgeAI:
I was experimenting just now and found something interesting. While the correct argument really is Error, sometimes a System.Exception will be passed there, which does not inherit from Error.
In fact as I tested more, I believe the type of the last argument should be object, because JavaScript will pass absolutely anything that you throw. You can even do:

```js
throw 5;
```

...and the last argument of your onerror handler will be set to 5.

LeonidVeriga:
Yes, a nice catch!

So that an error handler would look like:
```
void ErrorHandler(string message, string url, int lineNumber, int columnNumber, object error)
{
    Console.WriteLine("message: " + message);
    Console.WriteLine("url: " + url);
    Console.WriteLine("lineNumber: " + lineNumber);
    Console.WriteLine("columnNumber: " + columnNumber);
    Console.WriteLine("error: " + error);

    Error er = error as Error;
    if (er != null)
    {
        Console.WriteLine("As Error: " + er.Message);
    }

    Exception ex = error as Exception;
    if (ex != null)
    {
        Console.WriteLine("As Exception: " + ex.Message);
    }

    object o = error;
    if (o != null)
    {
        Console.WriteLine("As Object: " + o.ToString());
    }
}
```

TriEdgeAI:
Exactly. ;)

LeonidVeriga:
### Summary of changes

Changed `Action<Event> OnError` to `ErrorEventHandler OnError` for the following classes:
- `HTMLBodyElement `
- `Global`
- `Window`
- `WindowInstance`

Changed `ErrorEventHandler` delegate to have `error` parameter of `object` type:
```cs
public delegate bool ErrorEventHandler(string message, string url, int lineNumber, int columnNumber, object error);
```

PR #2057, opened 2016-11-11
[1606] v15.5 DateTime - default date should have year 1

### LINKED ISSUE #1606: DateTime - default date should have year 1
### Expected

`1/1/0001 12:00:00 AM`
### Actual

`now` or `1971` year
### Steps To Reproduce

``` csharp
public class App
{
    public static void Main()
    {
        var dt1 = new DateTime();
        Assert.AreEqual(dt1.Year, 1);

        var dt2 = default(DateTime);
        Assert.AreEqual(dt2.Year, 1);

        var dt3 = Activator.CreateInstance<DateTime>();
        Assert.AreEqual(dt3.Year, 1);
    }
}
```

Saltarelle also has this defect.

vladsch:
Fixes #1606

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -7291,8 +7291,29 @@
     });
 
     var date = {
+        offset: 62135596800000,
+
+        fromTicks: function (value) {
+            if (System.Int64.is64Bit(value)) {
+                value = value.div(10000).toNumber();
+            } else {
+                value = value / 10000;
+            }
+
+            return new Date(value - Bridge.Date.offset);
+        },
+
+        getTicks: function(dt) {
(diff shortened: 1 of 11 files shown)


LeonidVeriga:
Putting on-hold. It might be considered as a breaking-change and moved into **16.0**

TriEdgeAI:
I think the DateTime(long value) constructor will be incorrect even after this pull request. If we pass 0 as value it should take us to year 1 as well.
On another note, that constructor currently performs the division after extracting the integer value from the Int64 object, which might be problematic when dealing with very big numbers. Shouldn't the division be performed before extracting the number?

LeonidVeriga:
A new PR created for v16 branch - #2098

Issue #2065, opened 2016-11-12
Boxed enum types are treated as regular numbers


TriEdgeAI:
### Expected

```
Boat
Boat
Boat
```

### Actual

```
Boat
2
2
```

### Steps To Reproduce

[Deck](http://deck.net/50c6936a439e6c4dcafdb53edfce9dee)

```cs
public enum VehicleType
{
    Car,
    Plane,
    Boat
}

public class Program
{
    public static void Main()
    {
        VehicleType vehicleType = VehicleType.Boat;
        object box = vehicleType;
        
        Console.WriteLine(vehicleType);
        Console.WriteLine(box);
        Console.WriteLine(Enum.Parse(typeof(VehicleType), "Boat"));
    }
}
```

michaelcheers:
Related to #1347, #1317, #1312.

geoffreymcgill:
Boxing/Unboxing support has been implemented, but moving release to the 16.0 milestone.

michaelcheers:
Duplicate of [this comment](https://github.com/bridgedotnet/Bridge/issues/1309#issuecomment-217610825).

Issue #2055, opened 2016-11-11
GetType() returns wrong type due to lack of proper boxing


TriEdgeAI:
### Expected

```js
Bridge.assembly("Demo", function ($asm, globals) {
    "use strict";

    Bridge.define("Demo.Program", {
        $main: function () {
            var i = 0;
            var f = 0.0;

            Bridge.Console.log(Bridge.Reflection.getTypeFullName(System.Int32));
            Bridge.Console.log(Bridge.Reflection.getTypeFullName(System.Single));
            Bridge.Console.log(Bridge.Reflection.getTypeFullName(System.Int32));
            Bridge.Console.log(Bridge.Reflection.getTypeFullName(System.Single));

            var oi = SomeMagicBoxingTrickOlalala(i, System.Int32);
            var of = SomeMagicBoxingTrickOlalala(f, System.Single);

            Bridge.Console.log(Bridge.Reflection.getTypeFullName(Bridge.getType(oi)));
            Bridge.Console.log(Bridge.Reflection.getTypeFullName(Bridge.getType(of)));
        }
    });
});
```

```
System.Int32
System.Single
System.Int32
System.Single
System.Int32
System.Single
```

### Actual

```js
Bridge.assembly("Demo", function ($asm, globals) {
    "use strict";

    Bridge.define("Demo.Program", {
        $main: function () {
            var i = 0;
            var f = 0.0;

            Bridge.Console.log(Bridge.Reflection.getTypeFullName(System.Int32));
            Bridge.Console.log(Bridge.Reflection.getTypeFullName(System.Single));
            Bridge.Console.log(Bridge.Reflection.getTypeFullName(System.Int32));
            Bridge.Console.log(Bridge.Reflection.getTypeFullName(System.Single));

            var oi = i;
            var of = f;

            Bridge.Console.log(Bridge.Reflection.getTypeFullName(Bridge.getType(oi)));
            Bridge.Console.log(Bridge.Reflection.getTypeFullName(Bridge.getType(of))); //System.Int32 - WRONG
        }
    });
});
```

```
System.Int32
System.Single
System.Int32
System.Single
System.Int32
System.Int32
```

### Steps To Reproduce

[Deck](http://deck.net/690c6a3cb1929e3cc16244417e983bc9)

```cs
public class Program
{
    public static void Main()
    {
        int i = 0;
        float f = 0.0f;

        Console.WriteLine(typeof(int).FullName); 
[...]

Issue #1357, opened 2016-05-15
Boxed value types and String(s) should not be equal


michaelcheers:
### Expected

False
### Actual

True
### Steps To Reproduce

Type k and k again.
[Live Bridge](http://live.bridge.net/#8f3d57ec2ef3fa522eb5aea4422eaa2a)
[Dot Net Fiddle](https://dotnetfiddle.net/qJYsWz)

``` csharp
public class Program
{
    static object TryRead ()
    {
        string readline = Console.ReadLine();
        int value;
        if (int.TryParse(readline, out value))
            return value;
        else
            return readline;
    }
    [Ready]
    public static void Main()
    {
        object a = TryRead();
        object b = TryRead();
        Console.WriteLine(a == b);
    }
}
```

michaelcheers:
Also, if you type '7' then '7' again it has the same problem.

LeonidVeriga:
Yes, @michaelcheers, thank you, we reproduced the problem.

I just simplified you example:
[Live Bridge](http://live.bridge.net/#d98afb975c5e834b6acbb1d3cd1ce120)
[Dot Net Fiddle](https://dotnetfiddle.net/V0f7HE)
```
public class Program
{
    [Ready]
    public static void Main()
    {
        object a = Convert.ToString(3);
        object b = Convert.ToString(3);

        var r = a == b;
        Console.WriteLine("Should be false for strings: " + r);

        object a1 = 7;
        object b1 = 7;

        var r1 = a1 == b1;
        Console.WriteLine("Should be false for int: " + r1);
    }
}
```

weltkante:
> String (s) which are Object(s) should never be equal.

and

> Boxed value types and String(s) should never be equal.

This titles are misleading, they can be equal, just not in the sample code. It is easy to write code using boxed values or strings which are stored in object-variables and actually **must be** equal.

So the "never" in "should never be equal" is definitely wrong.

I adjusted the sample code to present an example where the comparison must be true.

```
using System;
					
public class Program
{
	public static void Main()
	{
        object a = Convert.ToString(3);
        object b = a;
		
		var r = a == b;
		
        Console.WriteLine("Should be true for strings: " + r);
		
		object a1 = 7;
        object b1 = a1;
		
		var r1 = a1 == b1;
		
        Console.WriteLine("Should be true for int: " + r1);
	}
}
```

weltkante:
[...]

Issue #1301, opened 2016-05-05
Ushort (s), short (s), uint (s), byte (s) give incorrect types


michaelcheers:
**Bridge Example** [Live](http://live.bridge.net/#088561871db47d9f14e55d02709001b7)
**.NET Fiddle Example** [Dot Net Fiddle](https://dotnetfiddle.net/SS4SJr)
JSIL has the same problem.

### Expected
```
System.Byte
System.UInt32
System.UInt16
System.Int16
```

### Actual
```
System.Int32
System.Int32
System.Int32
System.Int32
```

```cs
public class App
{
    public static void Main()
    {
        object v = (byte)3;
        Console.WriteLine(v.GetType().FullName);
        v = (uint)3;
        Console.WriteLine(v.GetType().FullName);
        v = (ushort)3;
        Console.WriteLine(v.GetType().FullName);
        v = (short)3;
        Console.WriteLine(v.GetType().FullName);
    }
}
```

michaelcheers:
So does [ILSpy](https://github.com/icsharpcode/ILSpy/issues/708).

PR #2078, opened 2016-11-18
[2077] Support boxing/unboxing operations

### LINKED ISSUE #2077: Support boxing/unboxing operations
https://msdn.microsoft.com/en-us/library/yz2be5wk.aspx
https://msdn.microsoft.com/en-us/library/ms228597.aspx
https://msdn.microsoft.com/en-us/library/aa691158(v=vs.71).aspx

### [Unbox(bool)] attribute
If class/method is marked as [External] then Bridge will unbox an argument automatically (if parameter type is object) because Bridge assumes that external libraries cannot work with boxed arguments. If such behaviour must be changed for some external classes/methods (if your external method can handle boxed arguments correctly) then apply attribute `Unbox(false)`

### Format of boxed value
```
{
   $boxed: true,
   v: value,
   type: T,
   constructor: T,
   getHashCode: function(),
   equals: function (o),
   valueOf: function(),
   toString: function ()
}
```

Related issues:
- #2077 Support boxing/unboxing operations
- #2065 Boxed enum types are treated as regular numbers
- #2055 GetType() returns wrong type due to lack of proper boxing
- #1514 referenceEquals is used when comparing strings
- #1357 Boxed value types and String(s) should not be equal (numbers only)
- #1347 All base types give incorrect type names
- #1317 String (+) operator doesn't work on Enums
- #1312 string.Format gives enum number(s) not names
- #1309 Booleans stored as objects give incorrect values
- #1301 Ushort (s), short (s), uint (s), byte (s) give incorrect types
- #1292 Char (s) give incorrect types when stored as objects.
- #1290 Chars don't give correct ToString when stored as object
- #1248 Doubles marked as objects don't give good types

vladsch:
Addresses:
- #2077 Support boxing/unboxing operations
- #2065 Boxed enum types are treated as regular numbers
- #2055 GetType() returns wrong type due to lack of proper boxing
- #1357 Boxed value types and String(s) should not be equal (numbers only)
- #1292 Char (s) give incorrect types when stored as objects.
- #1248 Doubles marked as objects don't give good types
- #1301 Ushort (s), short (s), uint (s), byte (s) give incorrect types
- #1290 Chars don't give correct ToString when stored as object

If class/method is marked as external then Bridge will unbox an argument automatically (if parameter type is object) because Bridge assumes that external libraries cannot work with boxed arguments. If such behaviour must be changed for some external classes/methods (if your external method can handle boxed arguments correctly) then apply attribute `Unbox(false)`

Format of boxed value:
```
{
   $boxed: true,
   v: value,
   type: T,
   constructor: T,
   getHashCode: function(),
   equals: function (o),
   valueOf: function(),
   toString: function ()
}
```

### DIFF
--- a/Bridge/Attributes/UnboxAttribute.cs
+++ b/Bridge/Attributes/UnboxAttribute.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace Bridge
+{
+    [External]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class)]
+    [NonScriptable]
+    public class UnboxAttribute : Attribute
+    {
+        public extern UnboxAttribute(bool allow);
+    }
+}
--- a/Bridge/Bridge.csproj
+++ b/Bridge/Bridge.csproj
@@ -57,6 +57,7 @@
     <AdditionalExplicitAssemblyReferences />
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="Attributes\UnboxAttribute.cs" />
     <Compile Include="Attributes\GlobalMethodsAttribute.cs" />
     <Compile Include="Attributes\MixinAttribute.cs" />
     <Compile Include="Attributes\ImmutableAttribute.cs" />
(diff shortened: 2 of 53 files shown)


LeonidVeriga:
A new PR created for **v16** branch - #2097

Issue #2068, opened 2016-11-12
Type.GetGenericTypeDefinition returns null for List<>


TriEdgeAI:
### Expected

```
System.Collections.Generic.List`1[T]
```

### Actual

```
null
```

### Steps To Reproduce

[Deck](http://deck.net/eeb218ada0cd3cf432c954ee781adf9c)

```cs
public class Program
{
    public static void Main()
    {
        Console.WriteLine(typeof(List<>).GetGenericTypeDefinition());
    }
}
```

PR #2085, opened 2016-11-23
[2067] Type.GetGenericTypeDefinition does not throw an exception when…

### LINKED ISSUE #2067: Type.GetGenericTypeDefinition does not throw an exception when invoked on a non-generic type
### Expected

```
System.InvalidOperationException: This operation is only valid on generic types.
```

### Actual

```
null
```

### Steps To Reproduce

[Deck](http://deck.net/7d2a7aa067916030c71eb00aaee8c172)

```cs
public class Program
{
    public static void Main()
    {
        Console.WriteLine(typeof(int).GetGenericTypeDefinition());
    }
}
```

vladsch:
Fixes #2067

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -3053,7 +3053,15 @@
         },
 
         getGenericTypeDefinition: function (type) {
-            return type.$genericTypeDefinition || null;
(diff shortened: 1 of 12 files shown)


vladsch:
It fixes https://github.com/bridgedotnet/Bridge/issues/2068 also

PR #2115, opened 2016-11-29
[2116] Add implicit operators from Any to Union (and back)

ProductiveRage:
Add implicit operators from `Any` to` Union` (and back) to make migration to `Union` (as` Any` is now obsolete) easier to do one piece at a time. Addresses #2116.

I would like to start updating my projects and libraries to move away from `Any` to `Union` but it would be easier if I could update one project at a time and still have it build, which I could if `Any` and `Union` could be used interchangeably.


LeonidVeriga:
@ProductiveRage, yes, it makes sense. Thank you for the PR.

Issue #1566, opened 2016-08-01
Math - Log does not work


LeonidVeriga:
- [x] Missing [MSDN Math.Log10 Method (Double)](https://msdn.microsoft.com/en-us/library/system.math.log10%28v=vs.110%29.aspx)
  - Compatible with .Net [tests](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/Math.cs#L251)
- [x] Missing [MSDN Math.Log Method (Double, Double)](https://msdn.microsoft.com/en-us/library/system.math.log10%28v=vs.110%29.aspx)
  - Compatible with the .Net [implementation](https://referencesource.microsoft.com/#mscorlib/system/math.cs,506)
  - Compatible with .Net [tests](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/Math.cs#L239)
- [x] Fix `Math.Log(Double)` to be compatile with .Net [tests](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/Math.cs#L228) 
- [x] Remove non C# API `decimal Log(decimal x, decimal logBase)`
- [x] Remove non C# API `decimal Ln(decimal x)`

Additionally, we need to create `double` overload.
### Steps To Reproduce
[Deck](http://deck.net/8c744462b526afc5aac50e24f3996df1)
``` csharp
public class Program
{
    public static void Main()
    {
        TestMathLog10();
        TestMathLogWithBase();
        TestMathLog();
    }

   public static void TestMathLog10()
    {
        AssertAlmostEqual(0.477121254719662, Math.Log10(3.0));
        AssertAlmostEqual(double.NegativeInfinity, Math.Log10(0.0));
        AssertAlmostEqual(double.NaN, Math.Log10(-3.0));
        AssertAlmostEqual(double.NaN, Math.Log10(double.NaN));
        AssertAlmostEqual(double.PositiveInfinity, Math.Log10(double.PositiveInfinity));
        AssertAlmostEqual(double.NaN, Math.Log10(double.NegativeInfinity));
    }

    public static void TestMathLogWithBase()
    {
        AssertAlmostEqual(1.0, Math.Log(3.0, 3.0));
        AssertAlmostEqual(2.40217350273, Math.Log(14, 3.0));
        AssertAlmostEqual(double.NegativeInfinity, Math.Log(0.0, 3.0));
[...]

PR #2132, opened 2016-12-05
[1566] Add working and .NET-compatible implementation of Math.Log

Zaid-Ajaj:
Addresses issue #1566

- Remove Math.Log(decimal, decimal) 
    - Does not work, see [Deck](http://deck.net/c21ea83967c81ed00adfc97a76031422)
    - System.Math in .Net does not have such overload for decimal types
- Remove Math.Ln(decimal)
    - Does not exist in System.Math
    - It is an alias for Math.Log
- Add Math.Log(double, double) that is
    - Compatible with the implementation from .NET [(See code here)](https://referencesource.microsoft.com/#mscorlib/system/math.cs,506)
    - Compatible with tests from [HERE](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/Math.cs#L239)
 - Add Math.Log10(double)

Script used to replicate the .NET tests from [JS Bin](http://jsbin.com/mevofadine/edit?js,console)
I Still can't seem to find Bridge.Test.dll which is why I am not able to test these changes

### DIFF
--- a/Bridge/Resources/Math.js
+++ b/Bridge/Resources/Math.js
@@ -23,6 +23,50 @@
             return Math.round(n) / m;
         },
 
+        // according to this implementation https://referencesource.microsoft.com/#mscorlib/system/math.cs,506
+        logWithBase: function (x, newBase) { 
+
+            if (isNaN(x)) {
+                return x;
+            }
+
+            if (isNaN(newBase)) {
+                return newBase;
+            }
+
+            if (newBase === 1) {
+                return NaN
+            }
+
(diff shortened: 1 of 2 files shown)


LeonidVeriga:
Зф

LeonidVeriga:
Hi @Zaid-Ajaj,

Thank you for the Pull Request. It looks very well.

We appreciate your links to the .Net implementation and tests. It helps a lot.
As **Bridge.Test** is in progress, we created a separate PR #2164 to incorporate your changes and add client tests. The fix is to be included into **15.6** release.

Issue #2149, opened 2016-12-08
Support DateTime's ToShortDateString() and ToShortTimeString()


Zaid-Ajaj:
MSDN [ToShortDateString()](https://msdn.microsoft.com/en-us/library/system.datetime.toshortdatestring(v=vs.110).aspx)

MSDN  [ToShortTimeString()](https://msdn.microsoft.com/en-us/library/system.datetime.toshorttimestring(v=vs.110).aspx)

Example usage: [Dotnet fiddle](https://dotnetfiddle.net/i6X3jq)

Workaround: [Deck](http://deck.net/c21c80af358bf07038468b9554e67d95/three)

PR #2150, opened 2016-12-08
[2149] Added ToShortDateString() and ToShortTimeString()

Zaid-Ajaj:
Addresses #2149.

Add both methods using a template attribute with the workaround suggested in [Deck](http://deck.net/c21c80af358bf07038468b9554e67d95/three)

**Update** I seem to have mixed my commits from the other pull request (still new to branching, any ideas?)


geoffreymcgill:
Thanks for submitting this pull request. We are reviewing. 

Seems like some Math.Log revisions have been mixed into this pull request. Can you remove those revisions so only the DateTime revisions are present.

Zaid-Ajaj:
I reverted the files concerning the Math class issue to match what there is on the master branche from this repo (is this the right way to do it?)

LeonidVeriga:
@Zaid-Ajaj,

Thanks for the Pull Request. It is nice to include into upcoming **15.6** release.

Regarding https://github.com/bridgedotnet/Bridge/pull/2150#issuecomment-265823121, yes, if you commit a wrong file then you just revert the changes. It is goog you created a separate branch **Zaid-Ajaj:Issue2149** in your repo so that it does not affect your **master** branch.

As **Bridge.Test** is in progress, we created a separate PR #2165 to incorporate your changes and add client tests. The only thing was required is to correct the templates to support culture locales - commit dfe7b60.

Issue #2198, opened 2016-12-19
Simultaneous use of [Module] and [Init] attributes


AndreyChechel:
When the function marked with `[Init]` attribute is located in a module, it will be called using a wrong reference. It'd be better to call it outside the module initialization function.

There are also connected defects (in the expected code):
- `main` function can't be resolved. However, other name will work fine.
- `Bridge.Console.log` will throw an exception:
>Uncaught TypeError: Cannot read property 'getHashCode2' of undefined
    at ctor.findEntry (bridge.js:10247)
    at ctor.set (bridge.js:10299)
    at ctor.add (bridge.js:10328)
    at ctor.f1 (bridge.js:23682)
    at bridge.js:1201
    at bridge.js:1101
    at ctor.init (bridge.js:23437)
    at new ctor (bridge.js:23431)
    at Function.getInstance (bridge.js:23337)
    at Function.logBase (bridge.js:23344)

### Expected

```js
Bridge.assembly("Demo", function ($asm, globals) {
    "use strict";

    define("Module1", function () {
        var Module1 = { };
        Bridge.define("Demo.Program1", {
            $scope: Module1,
            $main: function () {
                Bridge.Console.log("Hello World!");
            }
        });

        Module1.Demo.Program1.main();
        return Module1;
    });

});
```

### Actual

```js
Bridge.assembly("Demo", function ($asm, globals) {
    "use strict";

define("Module1", function (_) {
    var exports = { };
        Bridge.define("Demo.Program1", {
            $scope: exports,
            $main: function () {
                Bridge.Console.log("Hello World!");
            }
        });

        Demo.Program1.main();
    return exports;
});

});
```

### Steps To Reproduce

http://deck.net/90a630d0f7983f8496c98864afec8903

```cs
[Module("Module1")]
public class Program1
{
    [Init]
    public static void Main()
    {
        Console.WriteLine("Hello World!");
    }
}
```

vladsch:
This issue is fixed in the following commit
https://github.com/bridgedotnet/Bridge/pull/2218/commits/2b8957c9b3435d7701ee10419ab215d87e0de74f

PR #2218, opened 2017-01-08
[2217] [2198] Module functionality redesign

### LINKED ISSUE #2217: Module functionality redesign
Redesign module functionality.

Bridge now supports **AMD**, **CommonJS**, **UMD** and **ES6** modules (**AMD** is default).

The `[Module]` attribute can be applied to a class or assembly (in this case all classes belonging to the assembly will be included in the module).

The `[Module]` attribute can accept module type and/or module name. Required module code depends on module type (for example, `define` method is generated for **AMD**).

By default, if application uses a module then Bridge will generate code to load that module (for example, using `require` method). So, all modules will be loaded during application startup.

If you want to load modules manually (for example, bundle it to one file and include it into html) then module loading can be disabled using `loader` object in **bridge.json**.

```js
loader: {
    // loading system, by default AMD
    type: "AMD | CommonJS | ES6"

    // by default, "require"
    loaderFunction : string

    // true to disable automatic modules loading
    manualLoading: bool

    // mask to define which exactly modules should be excluded from auto-loading
    manualLoadingMask: string
    
    // true to disable variables generation if module is disabled then variable is created
    // which will be initialized after loading that module manually if call `Module.Load`
    skipManualVariables: bool
}
```

Module can be loaded manually using `Module.Load`
```js
await Module.Load(typeof (AnotherLibrary.Class1));
AnotherLibrary.Class1.Method();
```

Bridge doesn't include libraries for loading modules, required libraries (like require.js) must be added manually.

### Tasks
 - [ ] Update [Global Configuration](http://bridge.net/docs/global-configuration/) docs
 - [ ] add description `manualLoadingMask` - `[!][regex:][*...][;]`

vladsch:
#2217 Module functionality redesign
#2198 Simultaneous use of [Module] and [Init] attributes


Zaid-Ajaj:
This look awesome! I have a quick question
If I understand it right, this makes it possible to generate/output the javascript using modules (commonjs, umd... etc.) but will it make it possible to `require("other-modules")`? Because as of now, when writing a definition library we assume the javascript is already bundled and does not need to be `require`-ed anymore.

vladsch:
Hi @Zaid-Ajaj 

By default, Bridge will generate module loading via `require` automatically if a module is used inside your project (you have to attach javascript loader library to yor html page only). So, any code which uses code from module will wrapped by `require` method (as callback)

If you want to disable automatic loading then just set it in bridge.json, like
```
loader: {
   manualLoading: true
}
```

In this case, Bridge will not generate `require` instruction.

PR #2267, opened 2017-01-20
[2133] [2268] Generic Task argument doesn't get emitted to typescript

### LINKED ISSUE #2133: Generic Task<T> argument doesn't get emitted to typescript
### Expected

```ts
declare module Demo {
    export interface Program {
    }
        prototype: Program;
        new (): Program;
        main(): System.Threading.Tasks.Task<string>;
    }
    var Program: ProgramFunc;
}
```

### Actual

```ts
declare module Demo {
    export interface Program {
    }
        prototype: Program;
        new (): Program;
        main(): System.Threading.Tasks.Task;
    }
    var Program: ProgramFunc;
}
```

### Steps To Reproduce

http://deck.net/429ce9ca32a9af22a2f80c307e4ec19e

```cs
public class Program
{
    public static Task<string> Main()
    {
        Console.WriteLine("Hello World!");
        return null;
    }
}
```

vladsch:
Fixes #2133 and #2268


vladsch:
> It's unclear to me how this will solve #2133 as [IgnoreGeneric] is not used in the original sample

That attribute is applied to generic Task type, so no forcing because it is already applied
https://github.com/bridgedotnet/Bridge/blob/master/Bridge/System/Threading/Tasks/Task.cs#L95

>Why can't the generic args just be emitted in .d.ts and ignored in .js?

Well, i guess that if IgnoreGeneric is applied then js and d.ts should not ignore that attribute by default
However, we can change default behaviour like you suggested

vladsch:
Regarding AllowInTypeScript property in IgnoreGeneric attribute: such property is required and Bridge cannot handle it automatically. For example, if there is external library with typescript definition file. In javascript (and definition also) file a type is not generic but in C# such type is defined as generic (for example, for type checking) with IgnoreGeneric attribute. If we will ignore generic type parameters for javascript only and emit it in our typescript definition then our defintion file will be incorrect because it will use non generic type (which is defined in external typescript definition library) as generic (typescript compilation will fail).

Therefore default behaviour for IgnoreGeneric must be skipping type parameters for js and ts until AllowInTypeScript property will not be defined

PR #2292, opened 2017-01-28
[2270] Task.Yield().

TriEdgeAI:


TriEdgeAI:
Do you need one to justify why I am adding this at all, or do you simply want to add it to the automated tests? Would a recursive method which breaks it's recursion with Task.Yield() and checks the callstack afterwards be sufficient?

vladsch:
> Do you need one to justify why I am adding this at all, or do you simply want to add it to the automated tests?

Both

> Would a recursive method which breaks it's recursion with Task.Yield() and checks the callstack afterwards be sufficient?

We will appreciate for it

TriEdgeAI:
Okay, but I am really busy at this time, so I will probably do it after some 2-3 weeks, is that fine?

LeonidVeriga:
https://github.com/bridgedotnet/Bridge/pull/2292/commits/36915a998e0e7ebcd8718b6b12a160234fe85d85#diff-05cf499b2250a97fb7a4286b2f55fb09R55

Should we support [.Net](https://msdn.microsoft.com/ru-ru/library/system.threading.tasks.task.yield(v=vs.110).aspx) signature? `public static YieldAwaitable Yield()`

LeonidVeriga:
@TriEdgeAI,

This change also requires:
* [YieldAwaitable](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.yieldawaitable(v=vs.110).aspx) Structure
* [YieldAwaitable.YieldAwaiter](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.yieldawaitable.yieldawaiter(v=vs.110).aspx) Structure
* A short sample demonstrating Task.Yield() usage is also appreciated (your recursion [idea](https://github.com/bridgedotnet/Bridge/pull/2292#issuecomment-276125615) is great)

Please feel free to submit a new Pull Request when you have [time](https://github.com/bridgedotnet/Bridge/pull/2292#issuecomment-276638116).

Issue #1306, opened 2016-05-06
JSON.Parse breaks assumptions about long


Suchiman:
### Expected

I can JSON deserialize ObjectLiterals and work with them.
### Actual

Deserialized ObjectLiterals containing longs fail when used in operations such as comparisons because they don't have member methods like `ne`
### Steps To Reproduce

http://deck.net/195d16ff0c8630adf40d438941b5a22f

``` csharp
public class App
{
    public static void Main()
    {
        var t1 = JSON.Parse<Test>("{\"Id\":5}");
        var t2 = JSON.Parse<Test>("{\"Id\":10}");
        var result = t1.Id != t2.Id;
        Global.Alert(result.ToString());
    }
}

[ObjectLiteral]
public class Test
{
    public long Id { get; set; }
}
```

geoffreymcgill:
Thank you for the report. We have reproduced the problem and are investigating a fix.

LeonidVeriga:
Use `Bridge.JSON.Serialize` and `Bridge.JSON.Deserialize`. See #2418 for more information.

Issue #98, opened 2015-03-31
JSON.Parse<T> does not work with object instances


Mimisss:
Once implemented it is good to review this KB article:
http://bridge.net/kb/work-with-data

```c#
Foo foo = JSON.Parse<Foo>((string)obj)
```

### See Also

* http://forums.bridge.net/forum/community/help/130
* http://forums.bridge.net/forum/community/help/179
* http://forums.bridge.net/forum/general/feature-requests/1107
* http://forums.bridge.net/forum/bridge-net-pro/bugs/1417 (sub-objects)
* #894 (duplicate)
* #2302

geoffreymcgill:
This should work with primitive types, such as string, int, bool, etc. 

Enhanced support for deserializing Custom Types is still to be implemented.

LeonidVeriga:
Use `Bridge.JSON.Serialize` and `Bridge.JSON.Deserialize`. See #2418 for more information.

geoffreymcgill:
All Json serialization/deserialization support has been moved to the new **Bridge.Newtonsoft.Json** package. 

> install-package bridge.newtonsoft.json -pre

The following Deck demonstrates a basic scenario:

https://deck.net/newtonsoft.json

Issue #2069, opened 2016-11-13
typeof(object) is transpiled as Object, instead of System.Object


TriEdgeAI:
http://deck.net/7d3d21643005cdefca884b570b211f63
```cs
public class Program
{
    public static void Main()
    {
        Console.WriteLine(typeof(object).FullName);
        Console.WriteLine(new object().GetType().FullName);
    }
}
```

PR #2302, opened 2017-01-29
[98] [1306] [1432] Added proper support for JSON serialization/deserialization.

TriEdgeAI:
Addresses:
- #98
- #1306 
- #1432


geoffreymcgill:
This is going to take some time for us to work out where this functionality can be added, but we cannot modify the expected js emitted for **JSON.Stringify** and **JSON.Parse** members.

In C#, calling `JSON.Stringify(p1)` should emit to js exactly `JSON.stringify(p1)`.

In C#, calling `JSON.Parse(some_json_string)` should emit to js exactly `JSON.parse(some_json_string)`.

http://deck.net/5d5bc8936d0219ad2df8b301a7490c4b

```C#
public class Program
{
    public static void Main()
    {
        var p1 = new Person();
        
        // Must emit exactly:
        // var json = JSON.stringify(p1);
        var json = JSON.Stringify(p1);
        
        Console.WriteLine(json);
        
        // Must emit exactly:
        // var p2 = JSON.parse(json);
        var p2 = JSON.Parse(json);
        
        Console.WriteLine(((Person)p2).Name);
    }
}

[ObjectLiteral(ObjectInitializationMode.DefaultValue)]
public class Person 
{
    public string Name { get; set; } = "Frank";
}
```

So, this enhanced JSON serialize/deserialize functionality needs to find another home in the API. Likely the **Bridge** namespace makes the most sense. Something like the following:

```C#
var p1 = new Person();

string json = Bridge.JSON.Serialize(p1);

var p2 = Bridge.JSON.Deserialize<Person>(json);
```

or, taking inspiration from the JSON.NET api:
```C#
var p1 = new Person();

string json = Bridge.JsonConvert.SerializeObject(p1);

var p2 = Bridge.JsonConvert.DeserializeObject<Person>(json);
```

I would tend side with the later option and be inspired by the Json.NET api, which most of us are already familiar with.

Feedback appreciated.

TriEdgeAI:
In that case the generic JSON.Parse<T> methods need to be removed, because they are almost certain to produce unwanted results and mislead someone.
I would personally favour JavaScriptSerializer, because it is part of .NET, as opposed to Newtonsoft Json.NET which is a third-party library.

geoffreymcgill:
> In that case the generic JSON.Parse<T> methods need to be removed, because they are almost certain to produce unwanted results and mislead someone.

Agreed.

> I would personally favour JavaScriptSerializer

Agreed.
I forgot **JavaScriptSerializer** even existed. Never used it. :)

TriEdgeAI:
On a second thought... ASP.NET MVC uses Json.NET for it's serialization/deserialization needs, so it might make more sense to have that one implemented instead in order to pair it more easily with a backend project. :)

geoffreymcgill:
My preference would be to compile JSON.NET with Bridge and create a full compat version in JavaScript. I've looked into it already, but not possible _right now_.

vladsch:
I tested the changes and it seems all works prety good
The one issue which I noticed, Bridge already uses `System.String` as holder for string methods
https://github.com/bridgedotnet/Bridge/blob/master/Bridge/Resources/String.js#L484

And your pull request rewrites that holder. Need to use native "String" type still to avoid that overriding

TriEdgeAI:
Oops. Sorry. Should I fix it, or should we leave this altogether for a later time when you guys decide where to put this code exactly?

vladsch:
We will appreciate you if you update javascript code in the PR

TriEdgeAI:
Done.

geoffreymcgill:
My current thoughts are to move this functionality to **Bridge.JSON.Serialize** and **Bridge.JSON.Deserialize**.

TriEdgeAI:
Sounds good, but remove the generic JSON.Parse and make it return `object` so people don't get confused.

LeonidVeriga:
Yes, despite removing JSON.Parse<T> is a breaking change, it makes sense.
However, we would rather marked them with [Obsolete] then remove in next major release.

Bridge/Resources/JSON.js
```diff
@@ -0,0 +1,455 @@
+//TODO: Remove when issues are fixed.
+System.Object = Object; // https://github.com/bridgedotnet/Bridge/issues/2069
+System.Boolean = Boolean; // https://github.com/bridgedotnet/Bridge/issues/2062
+System.String = String; // https://github.com/bridgedotnet/Bridge/issues/2066
+System.DateTime = Date; // https://github.com/bridgedotnet/Bridge/issues/2064
```
geoffreymcgill: Issues #2064, #2066, #2062 and #2069 have been fixed and merged into **master**. If the latest **master** is merged into this pull request, these first five lines can be removed.

LeonidVeriga:
We created a new PR #2417 to have a separate branch within Bridge repo and address all changes based on the discussion above.

PR #2317, opened 2017-02-02
Create named functions for methods annotated with [Name] attribute

ProductiveRage:
Sorry for just throwing this at you out of the blue - it's just an idea and I thought I'd try my luck and see what you thought of it! (I've just remembered that your contributing guidelines demand that an issue be reported with you before a pull request is created.. hopefully you'll consider giving me a pass this time :)

There's an issue in the Bridge.React bindings about using Firefox with the React Dev Tools: it gives a view of what has been rendered by React and tries to attach names to each component. It's possible to create components from static methods and the Dev Tools try to use the name of the JavaScript function as the name for the rendered component. However, in the JavaScript that Bridge outputs, these functions are always anonymous and assigned to named properties - eg.

    myFunction: function () { /* .. */ ]

Some browsers are able to infer the name of the function by taking the name of the variable that it is assigned to (`myFunction` in this case), so the React Dev Tools work fine in Chrome (as Chrome supports this behaviour). However, Firefox does _not_ support it and so it shows the component in the Dev Tools as "No display name).

A workaround would be to name the function as well as naming the property that it is being assigned to - eg.

    myFunction: function myFunction() { /* .. */ ]

This repeating of `myFunction` feels a bit clumsy but I can't see any other way to make it work (short of waiting for Firefox to support this ES6 behaviour - but I've already been waiting for months now, I first recorded this issue in September, see https://github.com/ProductiveRage/Bridge.React/issues/13).

As a compromise, I thought that it might be better to only repeat the function name for cases where the name of the method is important in some way. I thought that perhaps only doing it when the `[Name]` attribute was applied to the method would be a good indicator.

Do you think that this makes sense?

### DIFF
--- a/Compiler/Translator/Emitter/Blocks/VisitorMethodBlock.cs
+++ b/Compiler/Translator/Emitter/Blocks/VisitorMethodBlock.cs
@@ -94,6 +94,8 @@ namespace Bridge.Translator
 
             if (hasNameAttribute)
             {
+				// If a method has a [Name] attribute then create a named function for this method property, rather than leaving it as an anonymous function (there are occasions that
+				// external libraries will use a function name for debugging purposes and the presence of a [Name] attribute seems like a good indicator that this might be the case)
                 this.Write(name);
             }
 
--- a/Compiler/Translator/Emitter/Blocks/VisitorMethodBlock.cs
+++ b/Compiler/Translator/Emitter/Blocks/VisitorMethodBlock.cs
@@ -29,6 +29,7 @@ namespace Bridge.Translator
 
         protected void VisitMethodDeclaration(MethodDeclaration methodDeclaration)
         {
+            var hasNameAttribute = false;
             foreach (var attrSection in methodDeclaration.Attributes)
             {
                 foreach (var attr in attrSection.Attributes)
@@ -60,6 +61,10 @@ namespace Bridge.Translator
                             return;
                         }
                     }
+                    else if (rr.Type.FullName == "Bridge.NameAttribute")
+                    {
+                        hasNameAttribute = true;
+                    }
                 }
             }
 
@@ -86,6 +91,11 @@ namespace Bridge.Translator
             this.WriteColon();
 
             this.WriteFunction();
+
+            if (hasNameAttribute)
+            {
+                this.Write(name);
+            }
 
             this.EmitMethodParameters(methodDeclaration.Parameters, methodDeclaration.TypeParameters.Count > 0 && Helpers.IsIgnoreGeneric(methodDeclaration, this.Emitter) ? null : methodDeclaration.TypeParameters, methodDeclaration);
 
--- a/Compiler/TranslatorTests/IntegrationTests/IntegrationTest.cs
+++ b/Compiler/TranslatorTests/IntegrationTests/IntegrationTest.cs
@@ -95,6 +95,7 @@ namespace Bridge.Translator.Tests
 #else
         [TestCase("19", true, true, TestName = "IntegrationTest 19 - Linked files feature #531 #562")]
 #endif
+        [TestCase("20", true, true, TestName = "IntegrationTest 20 - Test explicit naming of functions when [Name] attribute used")]
         public void Test(string folder, bool isToTranslate, bool useSpecialFileCompare, string markedContentFiles = null)
         {
             var logDir = Path.GetDirectoryName(FileHelper.GetExecutingAssemblyPath());
--- a/Compiler/TranslatorTests/TestProjects/20/Bridge/bridge.json
+++ b/Compiler/TranslatorTests/TestProjects/20/Bridge/bridge.json
@@ -1,0 +1,5 @@
+{
+  "output": "Bridge\\output",
+  "cleanOutputFolderBeforeBuildPattern": "*.*",
+  "outputFormatting": "Formatted"
+}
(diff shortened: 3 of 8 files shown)


geoffreymcgill:
@ProductiveRage – Thanks for submitting the pr. 

Just when `[Name]` is applied? Seems like this should apply to all functions?

Maybe could be a *bridge.json* and `[assembly: ]` level config to global turn off|on by developers. Other than increasing the final .js size, I don't really see a downside to adding this extra info.

ProductiveRage:
@geoffreymcgill Perhaps I was being pessimistic about how people might feel about this change! Considering it again, I think that enabling it globally would be excellent (otherwise there is a chance that it could seem unpredictable / undiscoverable; "why is this function getting named but this one being left anonymous?"). A config option to disable it could be a handy trap door..

About file size, I imagine that most applications that deliver Bridge JavaScript would have gzip'ing enabled and so I can't see it making a huge difference to the delivered file size (gzip is fantastic at efficiently transmitting repeated strings).

geoffreymcgill:
We will be reviewing this pull-request during our upcoming Monday team meeting.

geoffreymcgill:
We need a name for the setting to disable this feature in **bridge.json**. 

Current working name is **DisabledAnnotatedFunctionNames**, but it's a bit awkward and we're looking for other suggestions.

```json
{
    "disabledAnnotatedFunctionNames": true (false is default)
}

**Comment by LeonidVeriga** [comment]
Applying annotated function names does not seem to be an enabled-by-default feature.
However, the client tests JavaScript output for **Batch1** and **Batch3** increased by 2.2% and 1.9% respectively.

To enable this feature we suggest:
 - **bridge.json** setting **EnableAnnotatedFunctions** (false by default);
 - attribute **[AnnotatedFunction]** with targets Assembly, Class, Struct and Method. So that it can be enabled globally being applied on assembly or with a greater granularity by applying on a method or class.

**Comment by ProductiveRage** [comment]
Maybe the option could be called

  enableAnonymousClassMemberFunctions

or

  enableAnonymousClassMemberFunctionsSpaceOptimisation

?

I think that it _would_ be good for the function name to be included by default since it's helpful with the React debugging case and may well be useful for other developer tools. The only real downside is that it generates slightly more content (as @LeonidVeriga observes with the Batch1 / Batch2 sizes) - if this was a problem for someone then they might want to enable the "enableAnonymousClassMemberFunctionsSpaceOptimisation" option in your release configuration?

I would be interested how much larger the files are after they are gzip'd, since that is more likely to be relevant to how the files will be delivered (and I suspect that the increase will be even smaller than 2% since gzip is good at compressing duplicated textual terms).

**Comment by ProductiveRage** [comment]
I've had another thought about this and the possibility of the default-on behaviour (that the functions _should_ be named) - since this is only necessary because a significant browser (Firefox) hasn't yet supported some relatively new (ES6, apparently) functionality that other browsers (eg. Chrome) have, this should only be a temporary situation. For now, it could be an "always on" (or at least "default on") behaviour with the view that it would be disabled in the future when there are no significant browsers that require it any more?

**Comment by LeonidVeriga** [comment]
Yes, we think that this feature is just a workaround for debugging in Firefox at the moment. Since that it should default to switched-off (just **bridge.json** option **enableNamedFunctionExpressions**, no attribute) and to be explicitly enabled if required.

**Comment by ProductiveRage** [comment]
Because I'm losing track a little bit of what has been discussed and/or suggested and/or agree, would this option only apply to methods that have the [Name] attribute on them or would it apply to _all_ methods?

If you're currently thinking about a default-off option that applies to all methods, would you re-consider my original proposal that it default-_on_ but only apply to [Name] methods? (This would address my core use case while - hopefully - having minimal impact on others).

**Comment by geoffreymcgill** [comment]
It is hard for us to justify enabling this functionality by default, even if it is only added when `[Name]` attribute is applied.

Our current plan is rename the setting in **bridge.json** to `enableNamedFunctionExpressions`, default is `false`. 

The setting will only be available in **bridge.json** and would apply the extra function name to the whole assembly. It's an on|off setting, with an explicit decision to enable.

This allows developers to choose whether they want the extra name to be applied in their generated .js. This feature has no impact on the C# side and no downstream implications for inherited libraries. The only real penalty is maybe a 1-2% increase in .js file size, although as mentioned above, this may be mitigated with gzipping, but we haven't tested to confirm. 

As best we can determine, adding this extra function name is only required to work around a _current_ limitation of Firefox. At some point, I think it's fair to assume, that that limitation will fixed in Firefox, at which point there would be no use for this _feature_ in Bridge. 

We also do not want to enable this extra function name rendering for all libraries, and suffer the inevitable blow-back from those that do not understand the rational for adding the extra name. 

We also do not want to enable rendering this extra function name, then at some point (after Firefox supports) not render.

If a developer (or the library) requires this extra function name added to their generated .js, just explicitly set '"enableNamedFunctionExpressions": true` in your **bridge.json** and you're good to go. 

```js
{
    "enableNamedFunctionExpressions": true
}
```

LeonidVeriga:
We created a new PR #2317 to cerate a separate branch within Bridge repo and address all changes based on the discussion above.

LeonidVeriga:
**enableNamedFunctionExpressions** option changed in PR #2674: new name is **namedFunctions**.
See more info at https://github.com/bridgedotnet/Bridge/wiki/Global-Configuration#namedfunctions.

Issue #1717, opened 2016-08-29
Numeric, TimeSpan - Types properties. Interface declaration


LeonidVeriga:
### Expected

Type declaration missing interface declaration (result of `Type.GetInterfaces()` and `Type.IsClass`) for Numeric types:
- [x] `byte` (`Byte`)
- [x] `decimal` (`Decimal`)
- [x] `double` (`Double`)
- [x] `short` (`Int16`)
- [x] `int` (`Int32`)
- [x] `long` (`Int64`)
- [x] `sbyte` (`SByte`)
- [x] `float` (`Single`)
- [x] `ushort` (`UInt16`)
- [x] `uint` (`UInt32`)
- [x] `ulong` (`UInt64`)
- [x] `TimeSpan`

Affects 47 assertions.
P.S. The underlying functionality works, just missing the declaration.
### Steps To Reproduce

Sample for `Byte`

``` csharp
public class App
{
    public static void Main()
    {
        Assert.False(typeof(byte).IsClass);

        var interfaces = typeof(byte).GetInterfaces();
        Assert.AreEqual(interfaces.Length, 3);
        Assert.True(interfaces.Contains(typeof(IComparable<byte>)));
        Assert.True(interfaces.Contains(typeof(IEquatable<byte>)));
        Assert.True(interfaces.Contains(typeof(IFormattable)));
    }
}
```

LeonidVeriga:
Related to #15

Issue #1604, opened 2016-08-05
Date - creating instance through Activator should return todays date


LeonidVeriga:
### Expected

A todays date (like calling `new Date()`).
### Actual

A zero date of `1971` year.
### Steps To Reproduce

``` csharp
public class App
{
    public static void Main()
    {
        var fullYear = Activator.CreateInstance<Date>().GetFullYear();
        Assert.True(fullYear > 2011, fullYear + " > 2011");
    }
}
```

Issue #1567, opened 2016-08-01
Nullable - instance type name


LeonidVeriga:
### Steps To Reproduce

``` csharp
public class App
{
    public static void Main()
    {
        Assert.AreEqual(
            typeof(Nullable<double>).FullName,
            "System.Nullable$1[[System.Double, mscorlib]]",
            "Open FullName");

        Assert.AreEqual(
            typeof(int?).FullName,
            "System.Nullable$1[[System.Int32, mscorlib]]",
            "Instantiated FullName");
    }
}
```

michaelcheers:
@LeonidVeriga, your type full names aren't quite right, [see here](https://dotnetfiddle.net/i4zQLi).

LeonidVeriga:
@michaelcheers, there is some descrepancies

Issue #1559, opened 2016-08-01
TypedArray - should not allow Add, Remove, Clear


LeonidVeriga:
Related to #1548 Array - IList and ICollection methods chaning array should throw

### Actual

Convertions to generic versions of **IEnumerable**, **ICollection** and **IList** have  **Object** as a  subtype for generic type parameter.

`Float32Array`, `Float64Array`, `Int8Array`, `Int16Array`, `Int32Array`,  `Uint8Array`, `Uint16Array`, `Uint32Array` allows adding and removing elements through `IList` and `ICollection` interfaces.

`Add()`, `Insert()`, `Remove()` throws `NulReferenceException`.
`Clear()`  does not throw exception.

### Expected

Convertions to generic versions of **IEnumerable**, **ICollection** and **IList** should have a TypedArray subtype as a generic type parameter (for example, **float** for **Float32Array**).

All the methods below should throw `NotSupportedException`:
- `Add()`, `Insert()`, `Remove()`, `Clear()`
### Steps To Reproduce

``` csharp
public class App
{
        [Test]
        public void TypePropertiesAreCorrect_SPI_1559()
        {
            var interfaces = typeof(Float32Array).GetInterfaces();
            Assert.AreEqual(7, interfaces.Length, "Interface count should be 7");
            // #1559
            Assert.True(interfaces.Contains(typeof(IEnumerable<float>)), "Interfaces should contain IEnumerable<float>");
            Assert.True(interfaces.Contains(typeof(ICollection<float>)), "Interfaces should contain ICollection<float>");
            Assert.True(interfaces.Contains(typeof(IList<float>)), "Interfaces should contain IList<float>");

            object arr = new Float32Array(0);
            Assert.True(arr is Float32Array, "Is Float32Array");
            Assert.True(arr is IEnumerable<float>, "Is IEnumerable<float>");
            Assert.True(arr is ICollection<float>, "Is ICollection<float>");
            Assert.True(arr is IList<float>, "Is IList<float>");

        }

        [Test]
        public void ICollectionMethodsWork()
        {
            var coll = (ICollection<float>)new Float32Array(new float[] { 3, 6, 2, 9, 5 });
            Assert.AreEqual(coll.Count, 5, "Count");
            Assert.True(coll.Contains(6), "Contains(6)");
            Assert.False(coll.Contains(1), "Contains(1)");
            Assert.Throws<NotSupportedException>(() => coll.Add(2), "Add");
[...]

PR #2372, opened 2017-02-09
[2299] Calling Add on an array does not produce a runtime error.

### LINKED ISSUE #2299: Calling Add on an array does not produce a runtime error
### Expected
```
Run-time exception (line 9): Collection was of a fixed size.

Stack Trace:

[System.NotSupportedException: Collection was of a fixed size.]
  at Program.Main(): line 9
```
### Actual
```
2
```
### Steps To Reproduce
[Dot Net Fiddle](https://dotnetfiddle.net/tFNf6O)
[Live Bridge](http://deck.net/544eddd66d8d19638b3f00dbfba5b559)

```c#
public class Program
{
    public static void Main()
    {
        int[] array = new[] { 7 };
	((IList<int>)array).Add(5);

	Console.WriteLine(array.Length);
    }
}

vladsch:
- #2299 Calling Add on an array does not produce a runtime error
- #1548 [NDN] Array - IList and ICollection should be revised
- #1604 Date - creating instance through Activator should return todays date

Tests revised:
- #1559 TypedArray - should not allow Add, Remove, Clear
- #1567 Nullable - instance type name
- #1717 Numeric, TimeSpan - Types properties. Interface declaration


LeonidVeriga:
#2345 is not a part of this PR.

PR #2417, opened 2017-02-24
[1432] JSON serialization support

### LINKED ISSUE #1432: JSON.Parse<SomeType[]>(jsonText) returns array of plain JS objects instead of Bridge class

vladsch:
Fixes #1432

ObjectLiteral serialization/deserialization is supported partialy because there is no type definition and metadata for such types


LeonidVeriga:
@vladsch, check this use case http://deck.net/f38e2913b0ded8572c23c10a38b34870 (comment out property in INameEntity interface to get it working).

PR #2527, opened 2017-03-31
+ added overloaded methods for Math.Min\Max and other methods

dmitsov:
Fixes # .
There is an error when Math.Min\Max are given float arguments. One of our clients reported this bug to us.
- Add overloaded methods for several math functions in order to prevent errors when using the math functions with float arguments


dmitsov:
I apologise for the bad commit. I will do a new fork of the repository and try again

PR #2531, opened 2017-04-03
[2519] Two JavaScript files are generated in Folder (Lib) mode

### LINKED ISSUE #2519: Two JavaScript files are generated in Folder (Lib) mode
Two java script files are generated in Bridge/ourput using origin/master branch.
Problem appears when [Namespace(false)] is added

### Steps To Reproduce

https://deck.net/a9393901ad4e01678409a79707c54c93

```c#
[Namespace(false)]
public class MainClass
{
    [Ready]
    public static void Main()
        {
        }
}
```

### Expected Result

```js
Good question, what should be expected result. Probably only one file should be generated.
```

### Actual Result

```js
2 files are generated, one with project directory name and second with csproj name
```

LeonidVeriga:
Fixes #2519

### DIFF
--- a/Compiler/Builder/Program.cs
+++ b/Compiler/Builder/Program.cs
@@ -362,7 +362,20 @@ namespace Bridge.Builder
                     ? Path.GetFileNameWithoutExtension(bridgeOptions.ProjectLocation) : bridgeOptions.Folder;
             }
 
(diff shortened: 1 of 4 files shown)


LeonidVeriga:
@kcudnik, it would be great if you checkout this PR branch to check it solves the problem for your case.

PR #2532, opened 2017-04-03
System.math overloaded methods

dmitsov:
Fixes # .
Missing overloaded math methods for System.Math.Min\Max and a few other methods
- Overloaded methods in System.Math with float arguments. We use Bridge in our product and one of clients reported that they can't use System.Math.Min\Max with float arguments. We decided to add overloaded methods for Min\Max and some others just to be safe

### DIFF
--- a/Bridge/System/Math.cs
+++ b/Bridge/System/Math.cs
@@ -82,8 +82,8 @@ namespace System
         [Template("Bridge.Math.round({d}, 0, 6)")]
         public static extern double Round(double d);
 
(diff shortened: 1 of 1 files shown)


dmitsov:
Just to add. I made a previous pull request. But since it was my first one, I botched it. This is the clean version.

geoffreymcgill:
The **Math** requires a full review on both the C# and JavaScript sides. 

As well, there are errors in this PR that have not been corrected. We're closing this PR and the Bridge dev team will review the **Math** class(es).

PR #2533, opened 2017-04-03
[2537] WebSocket.SendAsync

DierkDroth:
https://forums.bridge.net/forum/bridge-net-pro/bugs/3900-websocket-sendasync-not-implemented-correctly


Dia6lo:
Everything looks pretty good, though there is minor formatting issue on offset and count all-in-one-line declaration.
As per @DierkDroth, it will be difficult to create a reliable unit test for this problem, though for simple demonstration wss://echo.websocket.org wil work just fine. Also, this issue is blocked by https://github.com/bridgedotnet/Bridge/pull/2534.

geoffreymcgill:
@Dia6lo – We're working on the **ArraySegment<T>** issue and will commit a fix soon. The pull request #2534 will likely be closed as we feel the fix lies elsewhere.

geoffreymcgill:
We don't have a sample to test this PR, so we're closing for now.

DierkDroth:
@geoffreymcgill I'm not sure I understand
a) the existing websocket code clearly is flawed (please correct me if I'm wrong)
b) my pull request did address the issue (to the best of my knowledge)
c) there are no existing websocket tests/samples
d) it unlikely is possible to create a websocket sample anyway using deck.net (since you needed client + server). The same applies to tests

Hence, why wouldn't you merge my pull request into MASTER or - even better - your team came up with a fix which addressed aspects (which I might have overlooked, please let me know) + tests/samples?

geoffreymcgill:
> or - even better - your team came up with a fix which addressed all aspects (which I might have overlooked, please let me know) + tests/samples?

We will be happy to take a look at this in the future, but right now the team is focused on other tasks.

DierkDroth:
I perfectly understand, which is why I took the initiative and addressed the issue for you. Hence, if bandwidth is the limiting factor, why not merge my pull request and your team - at a later time - could add whatever needed to be added there?

Dia6lo:
Example to the rescue
https://deck.net/a0da47aac5f428a286aa7401bb2fb5ff

Still, Deck and Nuget builds of Bridge don't have fix from #2593, so it will work only if you fix bridge.min.js manually.

DierkDroth:
Thanks @Dia6lo

DierkDroth:
@Dia6lo  is modified your sample to demonstrate the full scope of the problem.
https://dev.deck.net/832992bfc2b22f4f85a5cd15d7f095b5/three

geoffreymcgill:
Bridge 16.0.0-beta has been released and https://deck.net has been updated with this new release. 

Please confirm the following two questions:

1. Without PR #2533, rendering rows 1-5 which is incorrect?
2. The Deck below should only render two rows (`3` and `4`) to the console?
3. With PR #2533, the console only renders `3` and `4`?

```js
// Expected
3
4

// Actual
1
2
3
4
5
```

https://dev.deck.net/10e24fc18ce107bce85bcabe99360e0a

```csharp
using System;
using System.Net.WebSockets;
using System.Text;
using System.Threading;
using System.Threading.Tasks;

public class Program
{
    public static async void Main()
    {
        await Run();
    }

    private static async Task Run()
    {
        var client = new ClientWebSocket();
        
        await client.ConnectAsync(new Uri("wss://echo.websocket.org"), CancellationToken.None);
        
        var sendBytes = new byte[] { 1, 2, 3, 4, 5 };
        var sendBuffer = new ArraySegment<byte>(sendBytes, 2, 2);
        
        await client.SendAsync(sendBuffer, WebSocketMessageType.Binary, true, CancellationToken.None);
        
        var receiveBytes = new byte[sendBytes.Length];
        var receiveBuffer = new ArraySegment<byte>(receiveBytes, 0, receiveBytes.Length);
        var receiveResult = await client.ReceiveAsync(receiveBuffer, CancellationToken.None);

        // Should write '3' and '4'
        // '1' - '5' is wrong
      
        for (int i = 0; i < receiveResult.Count; i++)
        {
            Console.WriteLine(receiveBuffer.Array[receiveBuffer.Offset + i]);
        }
    }
}
```

DierkDroth:
on 1) confirmed
on 2) yes
on 3) confirmed

Thx

Issue #2477, opened 2017-03-20
New [Convention] Attribute


geoffreymcgill:
## [Convention] Attribute

Wiki [docs](https://github.com/bridgedotnet/Bridge/wiki/Attribute-Reference#convention).

Convention attributes can be _stacked_.

Replaces functionality from the old `preserveMemberCase` global config setting. See Issue #2476. 

#### Applies To

- assembly
- class
- interface
- property
- field
- method

## Constructors

```csharp
[Convention(Notation notation)]
[Convention(Notation notation, ConventionTarget target)]
```

## Properties

Name | Type | Description
---- | ---- | ----
Notation | Notation | Specifies notation to be applied by [Convention] attribute.
Target | ConventionTarget | Specifies target(s) to be filtered by [Convention] attribute.
Member | ConventionMember | Specifies type member(s) to be filtered by [Convention] attribute.
Accessibility | ConventionAccessibility | Specifies access modifiers to be filtered by [Convention] attribute.
Filter | string | Semicolon separated list of type paths (a type member's full name, for example) to be applied by [Convention] attribute.
Priority | int | Applied to assembly attributes only

#### Example

```csharp
// Emit all members (Props, Methods, Fields) in lowerCamelCase notation
[Convention(Target = ConventionTarget.Member, Notation = Notation.LowerCamelCase)]

// Emit only Properties as lowerCamelCase
[Convention(Member = ConventionMember.Property, Notation = Notation.LowerCamelCase)]

// Emit all public Fields as lowercase
[Convention(Accessibility = ConventionAccessibility.Public, Member = ConventionMember.Field, Notation = Notation.LowerCase)]
```

The `[Convention]` attribute can also be applied at the assembly level. The `[assembly: Convention]` can be added to any .cs within your project, such as your projects **/Properties/AssemblyInfo.cs** file. 

#### Example

```csharp
[assembly: Convention(Target = ConventionTarget.Member, Notation = Notation.LowerCamelCase)]
```

## Notation Enum

The following table details all **Notation** enum values and provides a sample output given a property with the identifier of `FirstName`.

Value | Emit
---- | ----
UpperCase | FIRSTNAME
LowerCase | firstname
LowerCamelCase | firstName
UpperCamelCase | FirstName

## ConventionTarget Enum

Value |
---- |
All
Class
Struct
Enum
Interface
ObjectLiteral
Anonymous
External
Member

## ConventionMember Enum

Value |
---- |
All
Method
Property
Field
[...]

Issue #2476, opened 2017-03-20
Remove preserveMemberCase global setting


geoffreymcgill:
The  `preserveMemberCase` global setting in **bridge.json** is being removed and replaced with a new `[Convention]` Attribute that can be applied to an `assembly`, `class`, `interface`, or member level . 

Please see Issue #2477 for details on **ConventionAttribute**.

### Old

```json
{
    "preserveMemberCase": true
}
```

### New

```csharp
[assembly: Convention(Notation.LowerCamelCase)]
```

## TODO

- [x] Remove **preserveMemberCase** from [documentation](https://github.com/bridgedotnet/Bridge/wiki/global-configuration#preservemembercase)
- [x] Close open issues related to **preservemembercase**
  - [x] #2036
  - [x] #1710
  - [x] #1192
  - [x] #819

Issue #2265, opened 2017-01-20
Mark items with [Obsolete(error = true)]


LeonidVeriga:
- [x] `[FieldProperty]` attribute is no longer required (see #2478)
In v**17.0** `[FieldProperty]` will be removed (see #2236)
- [x] `List<T>.GetRange()` method
In v**16.0** `List<T>.GetRange()` marked with Obsolete attribute (#2255).
In v**17.0** `List<T>.GetRange()` will be removed (see #2236)
- [x] `[Name(bool)]` attribute is no longer required (see #2477 New `[Convention]` attribute)

PR #2547, opened 2017-04-07
[2477] [2476] Add Convention attribute and remove PreserveMemberCase global config

vladsch:
Issue #2476, #2477 (#2265).

- **preserveMemberCase** global option is removed
- **[Convention]** attribute is introduced
- marked **[Name(bool)]** as obsolete with error message


Bridge/Attributes/Convention.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
+
+namespace Bridge
+{
+    /// <summary>
+	/// Can be applied to a member to change case notation in the compiled script.
```
geoffreymcgill: @vladsch – Can you fix the indenting here

Issue #2553, opened 2017-04-09
Implement string[] Environment.GetCommandLineArgs()


TriEdgeAI:
Maybe return an empty array just for the sake of easier cross-compiling? Or maybe return an array with GET arguments?

## See Also

- #2224

LeonidVeriga:
@TriEdgeAI, we submitted PR #2572 to support **Environment** class. That would be great if you could have a look at the changes proposed. Your feedback is appreciated.

Issue #2224, opened 2017-01-10
Support for Environment class


geoffreymcgill:
Currently support for **Environment** is spotty. The whole class should be reviewed.

**Methods**
https://msdn.microsoft.com/en-us/library/system.environment_methods(v=vs.110).aspx

**Properties**
https://msdn.microsoft.com/en-us/library/system.environment_properties(v=vs.110).aspx

## See Also

* #2553

PR #2572, opened 2017-04-15
[2224] Support for Environment class

LeonidVeriga:
Addresses #2224, #2553

This is not an exact copy of .Net behavior:
 - Enrironment variables are stored in **Environment** static field. I.e. it does not contain OS environment variables. However the API allows adding, storing, handling (for example, **public static string ExpandEnvironmentVariables(string name)**) and removing in .Net-like way;
- Things related to **command line** - `location.pathname` and `location.search`;
- **CurrentDirectory** - `location.pathname`
- **ExitCode** is stored as a property, allows setting by the property and by `Exit(int exitCode)`;
- **Is64BitOperatingSystem** tries to get it analizing `(n.userAgent.indexOf("WOW64") != -1 || n.userAgent.indexOf("Win64") != -1))`
- **ProcessorCount** - `navigator.hardwareConcurrency`
- **StackTrace** - an apporach like `(new Error()).stack`
- **TickCount** - `Date.now()`;
- **Version** returns Bridge Compiler version;
- **FailFast** just throws an exception (i.e. not a .Net behavior as a **finally** will be executed);
- the rest is just templates returning `""`, `false` etc.

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -11587,6 +11587,8 @@ Bridge.Class.addExtend(System.Boolean, [System.IComparable$1(System.Boolean), Sy
                 ]
             },
 
+            noKeyCheck: false,
+
             ctor: function (obj, comparer) {
                 this.$initialize();
                 this.comparer = comparer || System.Collections.Generic.EqualityComparer$1(TKey).def;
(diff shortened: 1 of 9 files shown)


TriEdgeAI:
Looks great! My only concern would be that `Environment.TickCount` is supposed to be an ever-growing timer, whereas `new Date()` could potentially give you a date which is earlier than one that you obtained by another call to `new Date()` a few seconds ago. Then again, there is no sane way to get an ever-growing timer in JS. You could increment something within a `setInterval()` but that's just overkill.

LeonidVeriga:
@TriEdgeAI, thanks for the feedback.

PR #2628, opened 2017-05-01
Enable compilation in Xamarin Studio

geoffreymcgill:


geoffreymcgill:
Will do this work in another PR.

PR #2629, opened 2017-05-01
Add preliminary System.Threading.Thread support using Web Workers

### LINKED ISSUE #1117: Support for System.Threading.Thread using workers.
[Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers)
[Threads](https://msdn.microsoft.com/en-us/library/system.threading.thread(v=vs.110).aspx)

Though there is the Thread class.
It has no constructor with more than one argument and I'm not sure if it's implemented at all.

retsimx:
Fixes #1117 .

- Add preliminary System.Threading.Thread functionality
- Add Worker Management and Worker bindings
- Add changes to handle bridge being loaded in a worker, so as not to run [Ready] functions.

Caveats and Details:
_Because of the architecture of Web Workers, it is impossible/difficult/inefficient to do some things, and some things don't work the way the C# api indicates._
- **Absolutely no shared access between workers or the main thread**, Threads can receive an object from the thread's Start function, and a thread may return an object, it is possible to send all data required/returned by threads using those objects. 
- ThreadStarts can only be public functions, class and lambda functions will not work, as there is no way pass the object reference through to the worker reliably, and there is no way to easily share the object between threads (see above)
- There are some issues with Serializing/Deserializing objects, so native types are most reliable at the moment. Maybe one day in the future when Bridge.Json.Serialize and Deserialize can handle recursive serialization, classes will be able to be used.
- The System.Threading.Thread constructor takes a list of strings representing the javascript files to be loaded in to the worker (Excluding Bridge.js which is managed internally)
- System.Threading.Thread::Start takes 3 parameters, the public thread entry point, the parameter to pass to the thread, and an onResult callback which is triggered when the thread finishes and returns it's result.
- Threads may "queue" starts, that is, you may call start multiple times with new parameters or even entry points and reuse the web worker, this was a design decision based around the length of time it takes to spawn a webworker, so this allows you to reuse a web worker without having to create it again and incur the setup time.
- Threads that queue starts must pass an onResult callback to the Start function, otherwise one one thread start may be queued at one time, and Thread::Result will be set to the result of the thread.
- Threads will set IsActive to false when there are no more thread starts in the queue and all running thread starts have completed.
- Threads have a Join function that does not match C#, due to the way message handling works between workers/the main thread, it is impossible to spin until the thread has finished as the message handler for the thread will never run while spinning and will cause a deadlock, this is alleviated by changing Thread::Join to have a callback that is triggered when the thread has finished. Threads with multiple thread starts in the queue will not Join until the queue is empty.
- Thread has a helper function _GetCurrentJsFileUri_ that can be used to get the fully qualified Uri of the javascript file that calls it, this is used internally by Thread to load Bridge in the worker, and can also be used in user code when calling the thread constructor in order to get the current javascript file.
- Threads have a fallback to running on the main thread in the event that the browser does not support Workers
- Threads can be used like workers, and can send and receive custom messages using OnMessage event handling, and PostMessage functions (Both part of Thread)
- **Only tested in Chrome and FireFox**

You can find a demo for finding all prime numbers within a range here (The demo now includes an example of custom message passing also): [https://github.com/retsimx/BridgeThreadingExample](https://github.com/retsimx/BridgeThreadingExample)

Live version: [https://retsimx.github.io/BridgeThreadingExample/](https://retsimx.github.io/BridgeThreadingExample/)
It may take 10 seconds or so to complete the initial baseline test in the main thread

Example output (on a 12 core i7-4930k):
```
Got message in worker:  hello from main thread:)
Got message in main thread:  hello from the worker thread :)
Starting next benchmark with 1 thread(s)...
Max number: 10000000, Threads: 1, Time taken: 5855ms, Number of primes: 664579
Starting next benchmark with 2 thread(s)...
Max number: 10000000, Threads: 2, Time taken: 3820ms, Number of primes: 664579
Starting next benchmark with 4 thread(s)...
Max number: 10000000, Threads: 4, Time taken: 2167ms, Number of primes: 664579
Starting next benchmark with 6 thread(s)...
Max number: 10000000, Threads: 6, Time taken: 1609ms, Number of primes: 664579
Starting next benchmark with 8 thread(s)...
Max number: 10000000, Threads: 8, Time taken: 1371ms, Number of primes: 664579
Starting next benchmark with 12 thread(s)...
Max number: 10000000, Threads: 12, Time taken: 1462ms, Number of primes: 664579
Starting next benchmark with 16 thread(s)...
Max number: 10000000, Threads: 16, Time taken: 1715ms, Number of primes: 664579
Starting baseline benchmark on main thread...
Max number: 10000000, Threads: (Main Thread), Time taken: 6168ms, Number of primes: 664579
Complete.
```

_It is expected that 1 thread is slower than on the main thread due to time to create the worker and communicate parameters and results_


geoffreymcgill:
Wow 😮 
Thanks!

We'll get right on this.

DanTup:
Is the goal here compatibility with existing C# code, or is this intended to be the web worker API for Bridge?

If you're building a new web app that uses web workers it would feel a little unnatural to have to do it via the C# Threading API.

retsimx:
That is definately true, would it be better to move it out of System.Threading.Thread? I don't think it will ever be possible to match C# compatibility exactly. Perhaps a System.Threading.WebWorker namespace?

geoffreymcgill:
Our goal would be to support as much of the **System.Threading** namespace as possible.

> would it be better to move it in to System.Threading.WebWorker instead?

Unless it's part of native C# API, it should stay out of the **System** namespace.

retsimx:
Ok, so assuming we can not have a Thread class that exactly matches the C#, what is the best course of action? Does Bridge need a Worker API?

geoffreymcgill:
The goal would be to spend extra effort on trying to match the **System** namespace. Both in the API and the expected result. Where it's not possible to match, then we deal with those on a case-by-case basis.

If the API entity or member does not match with **System**, then it cannot be added to the Bridge **System** namespace.

retsimx:
I've been busy recently, I've had a bit more of a think about the issues between the limitations of javascript and the C# Thread spec. I suppose most things can be made to match, except Join() which must require a callback, as there is no way for the message handler to fire while in a spinning loop. How much of an issue will it be that there is no shared context at all? Can this be a written limitation? There will need to be an extension to the constructor (or at least an additional function) to load scripts other than bridge in to the web worker. Finally, I'm not sure how else to go about getting the return value from the web worker without returning a single object and setting some kind of value within the thread instance. I am open to thoughts, otherwise, if this seems too difficult or diverges too far from your vision for Bridge, I'm happy for this pull request to be rejected and to move it in to a separate bridge package specific to web workers as @DanTup, pointed out.

ps. What to do about writing unit tests? I attempted to do so from the beginning but Bridge.Test (required by the test projects) was unavailable.

fabriciomurta:
This is an interesting piece of functionality we been discussing a few days ago.

I believe if we want to match the most existing C# threading, then we could just call this "System.Threading.Tasks implemented using Web Workers", and try to match most of the interface as possible. For the behaviors that can't be mapped, then throw some exception stating so. In the end if we can't implement a very important feature of the C# version then this may be the case to give up web workers -- but as far as the C# mapping is concerned, an implementation not using web workers could also work here.

But if we want to leave it clear to the user they are using Web Workers i.e. produce Web Workers support on bridge (focus the client-side aspect, and not .NET interface) then as already said, no System.* namespace, it should be something from itself; two alternatives:
- Extending the Bridge.Html5 framework to support that
- Creating a new Bridge framework (you can see the [WebGL](https://github.com/bridgedotnet/Frameworks/tree/master/WebGL) one for a base).

This alternate path lets you focus the JavaScript functionality for people looking to implement the javascript concept directly from bridge. The first approach (to focus C#) will mean changes to files within Bridge itself. Otherwise it is a framework/plugin development.

I believe a nice implementation of System.Threading.Tasks.Parallel can be done using web workers, albeit the limitations.

About the P.S. in the last comment, we've released Bridge.Test a few weeks ago, so you can now freely play around with unit tests with bridge!

geoffreymcgill:

PR #2681, opened 2017-05-15
[2657] DateTime.ParseExact throws exception when using "o" format specifier

### LINKED ISSUE #2657: DateTime.ParseExact throws exception when using "o" format specifier
An exception occurs when using `DateTime.ParseExact` with the `o` format specifier (`String does not contain a valid string representation of a date and time.`).

### Expectd Result
https://dotnetfiddle.net/ejH9mf

### Actual Result
JavaScript exception
```
System.Exception: Uncaught System.FormatException: String does not contain a valid string representation of a date and time.
Error: String does not contain a valid string representation of a date and time.
    at ctor.System.Exception.ctor (https://deck.net/resources/js/bridge/bridge.js?16.0.0-beta:5115:31)
    at new System.FormatException.ctor (https://deck.net/resources/js/bridge/bridge.js?16.0.0-beta:5435:35)
    at Function.System.DateTime.parseExact (https://deck.net/resources/js/bridge/bridge.js?16.0.0-beta:9217:27)
```

### Steps To Reproduce
https://deck.net/716ff807ec037b8e6dc85539414b80b2
```cs
var a = DateTime.UtcNow.ToString("o");
var b = DateTime.ParseExact(a, "o", CultureInfo.InvariantCulture);
Console.WriteLine(b.ToString("o"));
```

vladsch:
Fixes #2657


vladsch:
Javascript has only three digits in milliseconds.
Also, Bridge doesn't support DateKind

vladsch:
In .NET, that format uses 7 digits for milliseconds.
We can append zeros in Bridge.

PR #2771, opened 2017-06-10
[2396] Missing System.String.LastIndexOf Method overloads

### LINKED ISSUE #2396: Missing System.String.LastIndexOf Method overloads
https://msdn.microsoft.com/en-us/library/ms224421(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/ms224420(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/ms224422(v=vs.110).aspx

```c#
LastIndexOf(String, Int32, Int32, StringComparison)
LastIndexOf(String, Int32, StringComparison)
LastIndexOf(String, StringComparison)
```

michaelcheers:
Fixes #2396, I think it works but have no way to test it because the Bridge.Test
nuget package appears not to exist.


LeonidVeriga:
@michaelcheers The PR ostly looks good.

However we cannot move any further without unit tests added. Since the goal is to support .Net behavior then usual .Net tests are fine. You can either:
1) Provide a link into DNF snippet like this https://dotnetfiddle.net/3bV5t2
    Once the tesst are passing we are fine to include the tests with the proven expected results into Bridge.
2) After you have the test code in .Net you can add a new file with name **BridgeNNNN** (where **NNNN** is an issue number) into project **Batch3** (under the corresponding project folder depending on the issue number). See any existing Batch3's file for an example.

We would also ask you to add XML docs (summary, parameter description, result value description) for new types and/or type members (we usually get it from MSDN).

geoffreymcgill:

PR #2779, opened 2017-06-10
[2472] More Math.Abs overloads

### LINKED ISSUE #2472: More Math.Abs overloads
In Native C#, Math.Abs has [many more overloads](https://msdn.microsoft.com/en-us/library/system.math.abs(v=vs.110).aspx) to Bridge where there is only double.

michaelcheers:
Fixed #2472.


michaelcheers:
Can this be merged for this release, this would be very useful and the amount of changes are very small so it should be easy to review.

geoffreymcgill:
The pr looks good, but will likely have to wait for 16.1.

Bridge 16.0 has been finalised and the **-rc** release will be soon (maybe next week). We cannot add features to **Bridge 16.0.0-rc**, and only critical bugs will be fixed.

michaelcheers:
@geoffreymcgill, what is the time frame for 16.1 or at least a beta version of it?

michaelcheers:
@geoffreymcgill, I have done the changes requested.

LeonidVeriga:
@michaelcheers, I will not require correction of XML documents for this PR and will correct it. However, the comments you added will not appear in IntelliSence. The correct syntax is
```cs
        /// <summary>
        /// Summary here
        /// </summary>
        /// <param name="value">Description</param>
        /// <returns>Description here</returns>
        public extern int SomeMethod(string value);
```

michaelcheers:
@LeonidVeriga, weird, IntelliSense shows them on my machine. Anyway, I have changed the comments.

geoffreymcgill:

PR #2783, opened 2017-06-10
[1615] StringBuilder - missing substring constructor

### LINKED ISSUE #1615: StringBuilder - missing substring constructor
[MSDN](https://msdn.microsoft.com/en-us/library/y6zc419d%28v=vs.110%29.aspx)

> Initializes a new instance of the StringBuilder class from the specified substring and capacity.
> 
> ```
> public StringBuilder(
>     string value,
>     int startIndex,
>     int length,
>     int capacity
> )
> ```

### Steps To Reproduce
https://dotnetfiddle.net/RDIrZ4
https://deck.net/775025447ca2ec6fffc82f2838d128dc

``` csharp
public class App
{
    public void SubstringConstructorWorks()
    {
        var sb = new StringBuilder("some text", 5, 3, 55);
        Assert.AreEqual("tex", sb.ToString(), "Text");
        Assert.AreEqual(3, sb.Length, "Length");
    }
}
```

## See Also

* #2508

michaelcheers:
Fixed #1615.

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -9264,15 +9264,30 @@ Bridge.Class.addExtend(System.Boolean, [System.IComparable$1(System.Boolean), Sy
         ctor: function () {
             this.$initialize();
             this.buffer = [],
-            this.capacity = 16;
-
-            if (arguments.length === 1) {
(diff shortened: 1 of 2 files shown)


LeonidVeriga:
Fixed in #2908

PR #2789, opened 2017-06-11
[779] Add System.IO support

### LINKED ISSUE #779: Consider support/handling of System.IO
http://forums.bridge.net/forum/general/open-discussions/1320

> The System.IO namespace appears to be unimplemented. At least some basic support should be available as an extension package. Especially the exceptions like System.IO.TimeoutException should be available. While they are in the System.IO namespace they get used elsewhere.
> 
> In terms of the System.IO namespace, I think supporting Stream, Reader, Writer, and MemoryStream along with the readers/writers such as BinaryReader, BinaryWriter, TextReader, TextWriter etc would be sufficient. It would at least allow existing code that accepts a stream/reader/writer parameter. It occurs to me supporting the encoding classes would probably be necessary to help with the IO routines as well.

michaelcheers:
Fixes #779 (adds I/O support)

I've added 

- Stream
- MemoryStream
- BinaryWriter
- BinaryReader
- FileStream (made from Bridge.Html5.File)

I have tested everything except FileStream.


geoffreymcgill:
I'm going to close this Pull Request as **System.IO** is available as of Bridge 16.3.0.

PR #2790, opened 2017-06-11
[2634] Add String's ToLowerInvariant() and ToUpperInvariant() methods

### LINKED ISSUE #2634: Add String's ToLowerInvariant() and ToUpperInvariant() methods
I would like to use the methods [`ToLowerInvariant()`](https://docs.microsoft.com/en-us/dotnet/api/system.string.tolowerinvariant) or [`ToUpperInvariant()`](https://docs.microsoft.com/en-us/dotnet/api/system.string.toupperinvariant) to normalize string property values to ensure that the normalized values are the same independent of the local culture settings.
My use case is that I need to insert objects into a database that is case sensitive, but the case makes no difference to the user. So I would like to normalize string values before inserting them to lower case.

If I understood it correctly, Bridge only supports culture invariant methods of `ToLower()` and `ToUpper()`  anyway because specifying the culture is not possible in JavaScript. So could you additionally provide the methods `ToLowerInvariant()` or `ToUpperInvariant()` that simply call the methods `ToLower()` and `ToUpper()`? This would allow me to use `ToLowerInvariant()` in my C# code. Otherwise I would have to use `ToLower()` in C# which means that the normalization depends on the local culture settings.

### Steps To Reproduce

A minimal example with the C# code that I would like to use:

https://deck.net/372d0bcbaf5fa5be6ba0577efbcd3f0d

```c#
public class Person
{
    public string FirstName
    {
        get
        {
            return _firstName;
        }
        set
        {
            _firstName = value.ToLowerInvariant();
        }
    }
    private string _firstName;
}
```

### Expected Result

```js
    "use strict";

    Bridge.define("Demo.Person", {
        fields: {
            _firstName: null
        },
        props: {
            FirstName: {
                get: function () {
                    return this._firstName;
                },
                set: function (value) {
                    this._firstName = value.toLowerCase();
                }
            }
        }
    });
});
```

### Actual Result

I get a compile error because the `String` class has no method `ToLowerInvariant`.

michaelcheers:
Fixed #2634.


LeonidVeriga:
@michaelcheers, please see comment https://github.com/bridgedotnet/Bridge/pull/2771#issuecomment-321916248 regarding Unit tests.

geoffreymcgill:

PR #2791, opened 2017-06-11
[250] Allow option for Single-Quoted Strings in Output

### LINKED ISSUE #250: Allow option for Single-Quoted Strings in Output
It is a matter of personal preference. I would love it if there was some option to change the default quoting of strings to single instead of double.

michaelcheers:
Fixed #250.
Please check my changes carefully as I do not have very much experience with the Bridge emitter.

### DIFF
--- a/Compiler/Contract/Config/IAssemblyInfo.cs
+++ b/Compiler/Contract/Config/IAssemblyInfo.cs
@@ -198,5 +198,10 @@ namespace Bridge.Contract
         {
             get; set;
         }
+
+        char Quote
+        {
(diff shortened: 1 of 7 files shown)


geoffreymcgill:
https://github.com/bridgedotnet/Bridge/issues/250#issuecomment-312963695

geoffreymcgill:
We're going to keep things simple and standardise on double-quotes for strings. 

Maybe in the future, once the Bridge Plugin API has been finalised a **SingleQuotedStrings** plugin could be developed by the community.

PR #2813, opened 2017-06-17
[2185] More uri support.

### LINKED ISSUE #2185: Uri support
(this is a suggestion)

Currently there is only a stub of the Uri class, something like

```csharp
public class Uri
{
    public string AbsoluteUri { get; }

    public Uri(string uriString)
    {
        AbsoluteUri = uriString;
    }
}
```

I had to have a Uri support for Granular, so I've implemented most of its functionality as extension methods.
Unfortunately constructors cannot be extended in the same way, so a code that instantiate Uris couldn't be shared between .Net environment and Bridge environment.

It could be solved in Bridge by adding the 3 missing constructors (and exposing the given parameters as properties)
```
Uri(Uri baseUri, string relativeUri);
Uri(Uri baseUri, Uri relativeUri);
Uri(string uriString, UriKind uriKind);
```

It could also be solved by copying Granular's Uri implementation (I'm fine with it), and having a full support of Uri.

The Uri extension methods are implemented here
* [UriComponents.cs (parsing)](https://github.com/yuvaltz/Granular/blob/master/Granular.Common.Web/Compatibility/UriComponents.cs)
* [UriExtensions.cs ("properties")](https://github.com/yuvaltz/Granular/blob/master/Granular.Common.Web/Compatibility/UriExtensions.cs)
* [Uri.cs ("constructors")](https://github.com/yuvaltz/Granular/blob/master/Granular.Common.Web/Compatibility/Uri.cs)

Some unit tests are implemented here
* [UriTest.cs](https://github.com/yuvaltz/Granular/blob/master/Granular.Presentation.Tests/UriTest.cs)
(the tests are running on both .Net's Uri and the UriComponents above)

michaelcheers:
[2185] More uri support. Fixes #2185.


LeonidVeriga:
@michaelcheers, that look great.

The following things missing:
* xml doc for new types and type members
* tests... at least a link to .NET tests

michaelcheers:
@LeonidVeriga, for tests try [this](https://dotnetfiddle.net/7S9KHx). It is not complete. You could try to copy some unit tests from Granular.

michaelcheers:
@LeonidVeriga, now I look Granular more closely I think that these aren't the same methods as the c# ones?

geoffreymcgill:

PR #2819, opened 2017-06-18
[1586] allow + sign at the beginning of decimal.

michaelcheers:


Bridge/Resources/Decimal.js
```diff
@@ -19,11 +19,13 @@
                 v = v.replace(nfInfo.numberDecimalSeparator, ".");
             }
 
+            v = v.replace(/\s/g, "");
```
kcudnik: If you will call TryParse on decimal, and there is space in the middle of numbers in string, this replace will concatenate "1\<space\>1" to "11", and this TryParse will succeed, but should fail according to [fiddle](https://dotnetfiddle.net/QOBWZT).

michaelcheers: @kcundnik, Thanks, I will change it back.

Bridge/Resources/Decimal.js
```diff
@@ -19,6 +19,9 @@
                 v = v.replace(nfInfo.numberDecimalSeparator, ".");
             }
 
+            if (v[0] === '+')
```
kcudnik: This will not work if there will be space at the beginning of the string in TryParse, im actually curious why regex in line 22 will not handle "+" sign in the decimal number, regex seems to be correct?

michaelcheers: @kcudnik, I do not know regex so, you'll have to help me out on this.

kcudnik: Ahh, proble is not in this place, is in Decimal.js in part:
https://github.com/MikeMcl/decimal.js/blob/master/decimal.js#L4312
 isDecimal = /^(\d+(\.\d*)?|\.\d+)(e[+-]?\d+)?$/i,
this regex does not support + at the beginning, seems like a pull request for MikeMcl :)

kcudnik: Unfortunately MikeMcl will not fix this, but we can still fix this in bridge code, just place your "+" check and substr on line 27 after `v = v.replace(/\s/g, "");` then we will know that format is correct on line 22, and all spaces will be removed, so if + is there it will be first character and it can be safely removed.

michaelcheers: @kcudnik, OK done.

Bridge/Resources/Decimal.js
```diff
@@ -24,6 +24,9 @@
             }
 
             v = v.replace(/\s/g, "");
+
+            if (v[0] === '+')
```
geoffreymcgill: If you can update this to use `{ }` and `"` for strings, we'll review and write some tests for this pr. 

#### Example

```js
if (v[0] === "+") {
    v = v.substr(1);
}
```

geoffreymcgill:

PR #2821, opened 2017-06-18
[1585] Decimal - Failed to cast to ulong

### LINKED ISSUE #1585: Decimal - Failed to cast to ulong
### Expected
`0`

### Actual
`Arithmetic operation resulted in an overflow.`

### Steps To Reproduce

```csharp
public class App
{
    public static void Main()
    {
            decimal x = 0;
            Assert.AreEqual((ulong)(x - 0.9m), 0UL);
    }
}
```

michaelcheers:
Fixed #1585.

### DIFF
--- a/Bridge/Resources/Decimal.js
+++ b/Bridge/Resources/Decimal.js
@@ -262,6 +262,8 @@
             }
 
             if (tp === System.UInt64) {
+                if (v.gt(System.Decimal(-1)) && v.lt(System.Decimal(0)))
+                    return System.UInt64(0);
                 if (v.value.isNegative()) {
                     throw new System.OverflowException();
                 }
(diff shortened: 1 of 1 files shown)


Bridge/Resources/Decimal.js
```diff
@@ -262,6 +262,8 @@
             }
 
             if (tp === System.UInt64) {
+                if (v.gt(System.Decimal(-1)) && v.lt(System.Decimal(0)))
```
geoffreymcgill: Can you revise to include if block braces `{ }`?

#### Example

```csharp
if (v.gt(System.Decimal(-1)) && v.lt(System.Decimal(0))) {
    return System.UInt64(0);
}
```

geoffreymcgill:

PR #2822, opened 2017-06-18
[1583] Made decimal throw DivideByZeroException.

michaelcheers:


geoffreymcgill:

PR #2826, opened 2017-06-18
[1582] Nullable decimal casting to ulong or long does not create InvalidOperationException.

### LINKED ISSUE #1582: Decimal - No expected InvalidOperationException for long/ulong
### Expected
`System.InvalidOperationException`

### Steps To Reproduce

```csharp
public class App
{
    public static void Main()
    {
        Assert.Throws<InvalidOperationException>(() =>
        {
            var _ = (long)x2;
        });

        Assert.Throws<InvalidOperationException>(() =>
        {
            var _ = (ulong)x2;
        });
    }
}
```

michaelcheers:
Fixes #1582.

### DIFF
--- a/Bridge/Resources/Decimal.js
+++ b/Bridge/Resources/Decimal.js
@@ -242,6 +242,8 @@
     };
 
     System.Decimal.toInt = function (v, tp) {
+        if (v == null)
+            throw new System.InvalidOperationException("Nullable object must have a value.");
         if (!v) {
             return null;
         }


Bridge/Resources/Decimal.js
```diff
@@ -242,6 +242,8 @@
     };
 
     System.Decimal.toInt = function (v, tp) {
+        if (v == null)
```
geoffreymcgill: Hi @michaelcheers – Can you revise lines 245-249 to the following?

```csharp
if (v == null) {
    throw new System.InvalidOperationException("Nullable object must have a value.");
}

if (!v) {
    return null;
}
```

geoffreymcgill:

PR #2838, opened 2017-06-20
[2830] Array $set seems to be pointlessly cloning the input indexes (which makes it slow)

### LINKED ISSUE #2830: Array $set seems to be pointlessly cloning the input indexes (which makes it slow)
Doing some more performance profiling for TwistNTurn - I see that $set in Array.js is calling slice on its input indicies.  This causes an allocation while just trying to set a value in an array.  My javascript skills are weak, but the slice call does appear pointless at first glance...
I've tried removing it locally to good effect.

michaelcheers:
I'm not sure if this breaks anything but I think it just improves performance. Fixes #2830.


LeonidVeriga:
Nice change.

PR #2867, opened 2017-06-26
[1318] Add ArrayList.

michaelcheers:
Added ArrayList.


geoffreymcgill:
I'm going to close this since we'll only accept pr's in **System** that are using the original .NET native .cs source files.

We're hoping to support **ArrayList** in Bridge 17.0 using the compiled-from-source original **ArrayList.cs** file.

PR #2858, opened 2017-06-24
[962] Keys in a dictionary are returned in sorted order

michaelcheers:
Fixes #962 by compiling it from .NET.

PR #2891, opened 2017-06-29
Leverage JS engine object hash optimisations in the Dictionary when t…

ProductiveRage:
…he keys are strings and the default comparer is used

The Bridge Dictionary<,> implementation uses a fairly sensible approach of hashing the keys and then using those hash values as "buckets", in which multiple key/value pairs may be stored. When a new bucket is required then it may be added as a new property on a private "entries" object, taking advantage of the fact that JS treats everything as a hash table (and the fact that the JS engines are optimised to do so since it's such a commonly-used pattern). However, if string keys are used and the default comparer is being used (meaning that there are no string "normalisation" operations required - a case-insensitive string comparer would require some normalisation work, for example) then the work of hashing the keys in JavaScipt may be skipped and the keys themselves used directly as the hash table keys (while the JS engine may have to do some hashing of its own under the hood, this will be done in optimised native code, rather than in JavaScript). The longer that the keys are that are used, the more benefit that these alternate code paths provide.

This probably seems like a bit of a micro-optimisation! We were doing some performance investigation and, for the code that we were looking at, a lot of time seemed to be spent in getHashCode methods. We were using a custom dictionary key that essentially consisted of two strings. We changed it so that we concatenated the strings and then had strings as the dictionary keys, at which point we realised that we could avoid the hashing altogether if we were willing to use a JS object as a hash table instead of using the Dictionary class.

When we tried this out, the performance was improved. We could have left it there but I thought that it would be fairly easy to incorporate the changes into the Bridge Dictionary implementation by adding some special handling for cases where string keys were used. This would mean that we could continue to use the strongly-typed Dictionary in our code _and_ we could have the improved performance _and_ so could anyone else!

The following is a somewhat contrived example. To really see the benefit, you need long keys and you need many iterations. On my computer (in Chrome 59), the following reports around **3200ms** with the current Bridge code but **4ms** with the code that I'm submitting. With shorter strings, the performance gap shrinks but the proposed code should still always faster.

```
var dict = new Dictionary<string, int>();

var key = new String('x', 10000);
dict[key] = 123;

var timer = Stopwatch.StartNew();
for (var i = 0; i < 100000; i++)
{
	var f = dict[key];
}
timer.Stop();
Console.WriteLine($"Time taken: {timer.ElapsedMilliseconds}ms");
```

### DIFF
--- a/Bridge/Resources/Collections/Dictionary.js
+++ b/Bridge/Resources/Collections/Dictionary.js
@@ -90,6 +90,12 @@
                         this.add(name, obj[name]);
                     }
                 }
+
+                // If the keys are strings and the default comparer is being used (not one like that has to "normalise" strings, such as a case-insensitive comparer) then
+                // we can attach the key/value pairs directly to the entries object, which avoids having to hash the keys when storing or searching for them. This will mean
+                // that more work is done by the browser's native code (which is optimised for object property updates and retrievals) and less work needs to be done by
+                // JavaScript. The longer the string keys, the more beneficial this is (since the hashing cost is O(N) for strings).
+                this.isSimpleStringKey = (TKey === System.String) && (this.comparer === System.Collections.Generic.EqualityComparer$1(TKey).def);
             },
 
             containsPair: function(pair) {
@@ -117,10 +123,38 @@
             },
 
             getKeys: function () {
+                var e, keys;
+
+                if (this.isSimpleStringKey) {
+                    var keys = [];
+
+                    for (e in this.entries) {
+                        if (this.entries.hasOwnProperty(e)) {
+                            keys.push(e);
+                        }
+                    }
+
+                    return System.Array.init(keys, TKey);
+                }
+
                 return new (System.Collections.Generic.DictionaryCollection$1(TKey))(this, true);
             },
 
             getValues: function () {
+                var e, values;
+
+                if (this.isSimpleStringKey) {
+                    var values = [];
+
+                    for (e in this.entries) {
+                        if (this.entries.hasOwnProperty(e)) {
+                            values.push(this.entries[e].value);
+                        }
+                    }
+
+                    return System.Array.init(values, TValue);
+                }
+
                 return new (System.Collections.Generic.DictionaryCollection$1(TValue))(this, false);
             },
 
@@ -130,9 +164,16 @@
             },
 
             findEntry: function (key) {
-                var hash = this.comparer.getHashCode2(key),
-                    entries,
-                    i;
+                var hash, entries, i;
+
+                if (this.isSimpleStringKey) {
+                    if (this.entries.hasOwnProperty(key)) {
+                        return this.entries[key];
+                    }
+                    return;
+                }
+
+                hash = this.comparer.getHashCode2(key);
 
                 if (Bridge.isDefined(this.entries[hash])) {
                     entries = this.entries[hash];
@@ -154,7 +195,7 @@
 
                 for (e in this.entries) {
                     if (this.entries.hasOwnProperty(e)) {
-                        var entries = this.entries[e];
+                        var entries = this.isSimpleStringKey ? [this.entries[e]] : this.entries[e];
 
                         for (i = 0; i < entries.length; i++) {
                             if (this.comparer.equals2(entries[i].value, value)) {
@@ -198,13 +239,19 @@
                     return;
                 }
 
-                hash = this.comparer.getHashCode2(key);
                 entry = new (System.Collections.Generic.KeyValuePair$2(TKey, TValue))(key, value);
 
-                if (this.entries[hash]) {
-                    this.entries[hash].push(entry);
-                } else {
-                    this.entries[hash] = [entry];
+                if (this.isSimpleStringKey) {
+                    this.entries[key] = entry;
+                }
+                else {
+                    hash = this.comparer.getHashCode2(key);
+
+                    if (this.entries[hash]) {
+                        this.entries[hash].push(entry);
+                    } else {
+                        this.entries[hash] = [entry];
+                    }
                 }
 
                 this.count++;
@@ -223,9 +270,18 @@
             },
 
             remove: function (key) {
-                var hash = this.comparer.getHashCode2(key),
-                    entries,
-                    i;
+                var hash, entries, i;
+
+                if (this.isSimpleStringKey) {
+                    if (this.entries.hasOwnProperty(key)) {
+                        delete this.entries[key];
+                        this.count--;
+                        return true;
+                    }
+                    return false;
+                }
+
+                hash = this.comparer.getHashCode2(key);
 
                 if (!this.entries[hash]) {
                     return false;


DanTup:
@ProductiveRage Are there gains to be had by including integers here too (a really common use of dictionaries is fast lookups for integer-key entities)? They might already be returning themselves as hash keys so might not save much there but would eliminate the `equals` calls (which I presume is the reason for this change being a bit more than just skipping get `getHashCode` and using the raw string)?

ProductiveRage:
@DanTup I haven't measured but the gains are likely to be very small. With strings, there is the cost of hashing (which increases with string length) and avoiding that is the biggest win. With integers, the number itself will be used as the hash and so we don't save much. The also means that each bucket will only have a single value in and so there will only ever be zero or one equals checks to perform for integer lookups and the equals comparisons for numbers will be very fast.

So I don't expect that this special case makes much sense for anything except strings (anything else is either very cheap - like integers - or hard to generalise).

DanTup:
@ProductiveRage I figured the reason for this change being more than just skipping the `getHashCode` call was to also eliminate the (single, pointless) equals check. In my (possibly flawed?) testing, int comparisons are slower than "reasonable" sized strings:

```js
var start = performance.now();
for (i = 0; i < 100000; i++) {
    var equals = 100 == 500;
}
console.log(performance.now() - start);

start = performance.now();
for (i = 0; i < 100000; i++) {
    var equals = "this is a reasonable string for a dictionary key" == "this is a reasonable string for a dictionary key but different!";
}
console.log(performance.now() - start);
```

That said; these numbers are so low I'm not sure either are worth worrying about, so to there may be value in simplifying this change to only skip `getHashCode`, avoiding the other logic changes to keep the code simpler to understand?

ProductiveRage:
It wouldn't add much complexity to rename "isSimpleStringKey" to "isOptimisedKey" and to alter the logic from:

`
this.isSimpleStringKey = (TKey === System.String) && (this.comparer === System.Collections.Generic.EqualityComparer$1(TKey).def);
`

to something like:

`
this.isOptimisedKey = ((TKey === System.String) || (TKey === System.Int32)) && (this.comparer === System.Collections.Generic.EqualityComparer$1(TKey).def);
`

There won't be _as much_ benefit (particularly compared to very long string keys) but there is still some. So that you can try this for yourself, I've set up a Deck link which allows you to compare the current dictionary implementation to the proposed version (with a tweak to optimise for Int32 as well as String) here:

[deck.net/1da30778ef45a95608d6ada9cc1f183b](https://deck.net/1da30778ef45a95608d6ada9cc1f183b)

The first time that you let the code execute, your browser will hang for a few seconds since these are the result I get (in Chrome 59):

> String key test - Time taken for 'Standard': 3446ms
> String key test - Time taken for 'Optimised': 14ms
> Int key test - Time taken for 'Standard': 190ms
> Int key test - Time taken for 'Optimised': 13ms

So you can see that ints get some improvement but not on the same scale as strings.

DanTup:
Seems like I need to upgrade my machine!

> String key test - Time taken for 'Standard': 8281ms
> String key test - Time taken for 'Optimised': 34ms
> Int key test - Time taken for 'Standard': 536ms
> Int key test - Time taken for 'Optimised': 35ms

(to be fair, it's a 3-4 year old £200 Chromebook =))

Definitely not the same scale, reduce something to less than 10% is still a massive win IMO (and on mobile, time/CPU is battery)!

michaelcheers:
@ProductiveRage, I think #2858 might replace this pull request as it just copies the C# implementation.

DanTup:
@michaelcheers I think it's worth comparing the performance - this pull request is deliberately avoiding doing what C# does and takes advantage of raw javascript objects in order to skip `getHashCode` and `equals` calls.

Even if the current implementation is replaced in #2858 it may well be worth redoing this optimisation in that implementation.

ProductiveRage:
@michaelcheers @dantup correct on both counts, the from-source translation will be good but the performance won't be able to compare to this. It's similar with the Bridge regex vs native js regex, if you need raw performance in a tight loop then the browser implementations are a long long way ahead of js code. If we can get the best of both, that would be the best.

ProductiveRage:
@DanTup @michaelcheers @vladsch 

For completeness, I've prepared a Deck example that compares the "optimised dictionary" approach proposed by this PR to the translated-from-.NET-source version:

[deck.net/599ef79ca64505d525088ebb3a5875b7](https://deck.net/599ef79ca64505d525088ebb3a5875b7)

The results (on my computer, with Chrome) are as follows:

> String key test - Time taken for 'Translated': 3313ms
> String key test - Time taken for 'Optimised': 20ms
> Int key test - Time taken for 'Translated': 318ms
> Int key test - Time taken for 'Optimised': 26ms

Interesting to see that the translated dictionary seems to be faster than the current implementation for int keys (318ms vs 190ms) -

> String key test - Time taken for 'Standard': 3446ms
> String key test - Time taken for 'Optimised': 14ms
> Int key test - Time taken for 'Standard': 190ms
> Int key test - Time taken for 'Optimised': 13ms

However, the performance gap between the translated dictionary and the "optimised" dictionary for string and int keys is still very large.

I realise that PR 2858 is intended to match the behaviour of .NET in regards to returning keys in a particular order (as I understand it, it's not something that is guaranteed by the interface but it's a behaviour that people may be relying upon - rightly or wrongly), so there may be a little more work to combine the string-or-int-key-optimised version with the translated version (key enumeration order is not ensured for object property enumeration) such as maintaining a list of sorted keys (or sorting them when the Keys property is retrieved).

If this PR doesn't get accepted then I can change my code to use a custom lookup object rather than the Bridge Dictionary implementation - it's just that I thought that it might be beneficial to submit the change so that others may get the performance boost in the same scenarios (I suspect that string-key dictionaries are fairly commonly used).

michaelcheers:
@ProductiveRage, wow, that performance is significantly faster. I guess in this case Bridge is really really slow. @ProductiveRage, do the keys in your dictionary implementation return in sorted order because although it is incorrect to assume their order will be the same, a lot of C# code relies on this as I found while porting [SpaceConvergence](https://github.com/lauriecheers/SpaceConvergence). Especially, JSONTable's LRCEngine. So, I copied the Dictionary source from c# referencesource and changed it a bit and it worked (which is why I made the pull request). You can see [here](http://michael.cheersgames.com/Demo/).

ProductiveRage:
@michaelcheers No, unfortunately the keys will not be guaranteed to be ordered - just so that I'm 100% clear, are the keys in .NET returned sorted alphabetically (or whatever the comparer specifies) or are they returned in order of insertion?

It should be possible to do a bit of extra booking-keeping with the optimised dictionary to maintain this behaviour but it's not in the pull request currently.

DanTup:
By coincidence I just came across this:

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map

IE support is ropey (IE11 and only partial) but interestingly it says:

> Any value (both objects and primitive values) may be used as either a key or a value

and

> A Map object iterates its elements in insertion order 

Which makes me wonder whether it'd actually be better to map the whole of the Dictionary on to this class (where it's supported, and fallback to something else if old IE). This might give the performance boosts mentioned here without restricting to simple string/int keys? (this could also solve the niggle @ProductiveRage has about not preserving key order).

That said - in a video I watched only this weekend, I'm sure Lars Bak said in v8 they preserved the order of things in objects because although the spec didn't require it, all other browsers did it. Haven't tested it though!

DanTup:
> I realise that PR #2858 is intended to match the behaviour of .NET in regards to returning keys in a particular order (as I understand it, it's not something that is guaranteed by the interface but it's a behaviour that people may be relying upon - rightly or wrongly)

The docs say `The order in which the items are returned is undefined`. If anyone is relying on them being in an order I think they're on dodgy ground even just in .NET. I don't think Bridge should spend any effort on trying to reproduce the current .NET behaviour because it could very well be changed on a whim (in a future .NET version, or a different .NET runtime) and then it'd be wrong anyway?

ProductiveRage:
@DanTup - you could only use the Map for cases where the default equality comparer is used, if a custom implementation was specified then you would need a way to deal with that (basically, you would have to use a custom / translated-from-.NET dictionary implementation which would be slower than native).

ProductiveRage:
@michaelcheers I meant to pick up this comment of yours earlier:

> I guess in this case Bridge is really really slow

It's probably a bit unfair to say that Bridge is slow here, it's more that this use case is common in JavaScript and so the JS engines have been really highly optimised to deal with it!

DanTup:
> you could only use the Map for cases where the default equality comparer is used

Whoops, good point (again)! Still, it might be worth considering even in just this case, as it'd still cover more cases than the simple string-or-int case. Ofc I've done zero investigation into it, the implementation might not be trivial!

michaelcheers:
@DanTup, I think that javascript's implementation is in order of keys: see this [jsfiddle](https://jsfiddle.net/mqw534xa/). Anyway, I think we shouldn't rely on the implementation of particular browsers.

DanTup:
@michaelcheers 

> I think that javascript's implementation is in order of keys

Yeah, that's what I mentioned above. Lars Bak said v8 copied other browsers, though it's not actually in the spec. My point was more about whether Bridge should try to copy C# behaviour where it's specifically documented as undefined - I don't (personally) think that's a worthwhile goal. If Bridge can have a faster dictionary that behaves as the C# one is spec'd, but has some differences for behaviour that is documented as not defined then I think it should go with the faster one! (ofc, it's not my project and I don't have to support it ;-))

calcock:
There is a lot of discussion of performance, and it looks like only tests in Chrome have been made so far.  Using the deck.net sample (https://deck.net/599ef79ca64505d525088ebb3a5875b7) here are some results from 3 other major browsers along with a comparison run in Chrome, and a run through Node.js:

Chrome 59

> String key test - Time taken for 'Translated': 2994ms
> String key test - Time taken for 'Optimised': 20ms
> Int key test - Time taken for 'Translated': 310ms
> Int key test - Time taken for 'Optimised': 23ms

IE 11 (Win 10)

> String key test - Time taken for 'Translated': 34538ms
> String key test - Time taken for 'Optimised': 2137ms
> Int key test - Time taken for 'Translated': 842ms
> Int key test - Time taken for 'Optimised': 79ms

Firefox 54

> String key test - Time taken for 'Translated': 1856ms
> String key test - Time taken for 'Optimised': 1704ms
> Int key test - Time taken for 'Translated': 143ms
> Int key test - Time taken for 'Optimised': 7ms

Edge (Win 10, current)

> String key test - Time taken for 'Translated': 11220ms
> String key test - Time taken for 'Optimised': 2055ms
> Int key test - Time taken for 'Translated': 359ms
> Int key test - Time taken for 'Optimised': 22ms

Node.js (6.10.3, Win 10)

> String key test - Time taken for 'Translated': 954ms
> String key test - Time taken for 'Optimised': 30ms
> Int key test - Time taken for 'Translated': 266ms
> Int key test - Time taken for 'Optimised': 5ms 

(All tests performed on the same 6th Gen i7 Desktop)

Thankfully in all cases the 'Optimised' test is faster, however none of the others see the same ~100 times improvement for string tests seen in Chrome.  In IE and Edge the current performance is pretty hideous, but the optimised performance still isn't what I would consider great (i.e only equivalent to un-optimised Chrome).  

The result that is interesting is the String based test in Firefox - the optimised is only slightly faster than the un-optimised. 

In all the browsers the improvement scales linearly with number of iterations (as you would expect), but changes to the string length of the key seems to have more varied effects in the different browsers - although he optimised one seems to fare better (i.e. more linearly) in my limited tests.

I don't really have a conclusion here - just wanted to make sure that the gains across the board were better understood.

michaelcheers:
I have to say well done Google Chrome, it has the fastest string key time and Firefox has the fastest int key time, well node.js but that's not a browser. I have a suspicion, I think google chrome might have optimized the loop and figured out it was doing the same thing 100000 times and did it once because it is so ridiculously faster than all the other browsers. I think node.js might have done the same thing.

DanTup:
I'd expect nodejs would be the same as Chrome since it's based on v8. I don't think Chrome is "optimising it away" because if you change the number of iterations the time taken seems to scale linearly still as you'd expect.

LeonidVeriga:
PR #2946 is going to be mreged into **master** for **16.0** with the suggested simple types key performance optimization and addressing key ordering issue.

Many thanks for the suggestions!

PR #2910, opened 2017-07-03
[2890] Bad handling of file encoding for javascript files included in Bridge projects

### LINKED ISSUE #2890: Bad handling of file encoding for javascript files included in Bridge projects
We discovered a strange issue where we had some source files (CKEditor) that contained code like this:

```js
var something = "\u200b";
```

This is a zero width space. At some point (we suspect minificantion) these appear to be replaced with literal bytes for a zero-width space (this makes sense). However, something goes wrong with the encoding and the character is then replaced with a question mark `?` in the Bridge output.

While testing, we noticed that very strange things happen with file encodings, for example:

- Create two bridge projects, one that references the other
- Add the following content to two javascript files, saved with UTF8 and UTF16 encodings:
    ```js
    var zeroWidthSpaceEscapedUtf8File = "\u200b";
    var zeroWidthSpaceUtf8File = "​";
    ```
- Add this to bridge.json:
    ```js
    "resources": [
      {
        "name": "all.js",
        "files": [
          "utf8.js",
          "utf16.js"
        ]
      }
    ]
    ```
- Reference this project from the second project and then build
- Review the contents of `all.js` in the output folders

### all.js

```text
var zeroWidthSpaceEscapedUtf8File = "\u200b";
var zeroWidthSpaceUtf8File = "​";
v a r   z e r o W i d t h S p a c e E s c a p e d U t f 1 6 F i l e   =   " \ u 2 0 0 b " ; 
 
 v a r   z e r o W i d t h S p a c e U t f 1 6 F i l e   =   "  " ; 
```

### all.min.js

```text
var zeroWidthSpaceEscapedUtf8File="​",zeroWidthSpaceUtf8File="​";v;a;r;z;e;r;o;W;i;d;t;h;S;p;a;c;e;E;s;c;a;p;e;d;U;t;f;F;i;l;e=" \ u 2 0 0 b ";v;a;r;z;e;r;o;W;i;d;t;h;S;p;a;c;e;U;t;f;F;i;l;e="  ";
```

It seems that the encoding of the first file is used for all files being combined, so in this case the utf16 file (which has two-bytes-per-char) is being read as if it was utf8 (resulted in all the extra bytes).

I did also reproduce the issue with the zero-width space being replaced with a literal `?` while typing this up, but I somehow can't repro again now. I think there was something that resulted in an ascii output (I can't repro this now, even with an ascii file thrown in). However I expect fixing this issue would resolve that too.

LeonidVeriga:
Addresses #2890:

- If resource consists of one file and does not contain any header or remark then it is embedded in its original encoding (old behaviour);
- If resource has more than one file or header/remark then it is read in its original encoding (BOM checked) and being embedded in the default UTF8 encoding
- Reading resource header with BOM checking as well


Compiler/Translator/Translator/Translator.Resources.cs
```diff
@@ -763,13 +769,41 @@ private bool ReadResourseFiles(string outputPath, MemoryStream buffer, ResourceC
...
+                this.Log.Trace("Reading resource file " + file.FullName + " as one-file-resource");
+                content = File.ReadAllBytes(file.FullName);
+            }
+            else
+            {
+                this.Log.Trace("Reading resource file " + file.FullName + " via StreamReader with byte order mark detection option");
+
+                using (var m = new StreamReader(file.FullName, Translator.OutputEncoding, true))
+                {
+                    if (m.CurrentEncoding != OutputEncoding)
```
AndreyChechel: `m.CurrentEncoding` will always return the specified default encoding here since **StreamReader** needs to read at least one byte first to detect the actual encoding. It could be done using **Peek** method:
```cs
m.Peek();
if (m.CurrentEncoding != OutputEncoding)
/* ... */
```

PR #2958, opened 2017-07-12
File logger changes

LeonidVeriga:

- Fresh log file on each compile (ovewrites if exists)
- Fix log file location if it has a leading dot in folder path. For example, an output folder **Folder1/.folder2** is now considered as subfolder **.folder2** in folder **Folder1** (not a  file**.folder2** in folder **Folder1**)
- Fix logging a list of loaded assemblies (it was failing for dynamic assemblies);
- **CleanOutputFolderBeforeBuildPattern** bridge.json option to support exclusive mask. For example, the following setting will delete all files (`*.*`) in output folder except **myFile.txt**:
   ```
   // by default, setting cleanOutputFolderBeforeBuild deletes files using mask "*. js|*.d.ts"
   // cleanOutputFolderBeforeBuildPattern can override that default mask
   "cleanOutputFolderBeforeBuildPattern": "*.*|!myFile.txt"
   ```
 - Log file is not removed even if **cleanOutputFolderBeforeBuild** or **cleanOutputFolderBeforeBuildPattern** are enabled


geoffreymcgill:
Nice changes. Thanks for adding!

PR #3021, opened 2017-08-04
[1592] Numeric - TryParse should set zero if unsuccessfull

### LINKED ISSUE #1592: Numeric - TryParse should set zero if unsuccessfull
[MSDN](https://msdn.microsoft.com/en-us/library/tkktxbeh%28v=vs.110%29.aspx):

> When this method returns, contains the 8-bit unsigned integer value equivalent to the number contained in s if the conversion succeeded, or zero if the conversion failed. The conversion fails if the s parameter is null or String.Empty, is not of the correct format, or represents a number less than MinValue or greater than MaxValue. This parameter is passed uninitialized; any value originally supplied in result will be overwritten.

### Actual
If input string representation can be parsed to number, returns the value before checking type range.

### Steps To Reproduce
https://dotnetfiddle.net/eCkKkz
https://deck.net/184aaa2f2ec59c6072908213187ee69a
``` csharp
public class App
{
    public static void Main()
    {
        bool result;

        byte b;

        result = byte.TryParse("54768", out b);
        Assert.AreEqual(0, b);

        result = byte.TryParse("-1", out b);
        Assert.AreEqual(0, b);

        short s;

        result = short.TryParse("54768", out s);
        Assert.AreEqual(0, s);

        result = short.TryParse("-55678", out s);
        Assert.AreEqual(0, s);

        sbyte sb;

        result = sbyte.TryParse("54768", out sb);
        Assert.AreEqual(0, sb);

        ushort us;

        result = ushort.TryParse("-1", out us);
        Assert.AreEqual(0, us);

        uint ui;

        result = uint.TryParse("-1", out ui);
        Assert.AreEqual(0, ui);
    }
}
```

LeonidVeriga:
Fixes #1592 - this is a duplicate PR to accept PR #2776 from fork. The problem is that we do not allow building "external" PRs on CI servers (AppVeyor/Travis, to prevent security issues) and the current settings of **Bridge** repo does not allow merging PRs when required [status](https://developer.github.com/v3/repos/statuses/) checks not passing (_Required statuses must pass before merging_).


LeonidVeriga:
The PR had base **master** branch by mistake, it had been corrected in PR #3022

PR #3049, opened 2017-08-13
[3048] DateTime.ToString with "g" format different for "en-US" locale

geoffreymcgill:
Fixes # 3048

Several default DateTime formats were corrected


geoffreymcgill:
Wrong fix. Deleting branch Issue3048.

PR #3059, opened 2017-08-15
[V16.1] Correct file name casing for Integration Test reference files

LeonidVeriga:
Some Integration Test reference folders contain duplicated files: basically the same files but different name casing. For example, **TestProject.Issues.N2198.js** and **testProject.issues.n2198.js**:
gitfilenamecaseissue

This causes unresolving conflicts when merging branch **v16.1** into **master**. The PR is to remove _duplicate_ files in **v16.1** branch (see the same for **master** branch by PR #3058).


LeonidVeriga:
Result:
image

PR #3080, opened 2017-08-23
[3063] Clone of assignmentExpression.Left in Preconverter is changing runtime behavior

### LINKED ISSUE #3063: Clone of assignmentExpression.Left in Preconverter is changing runtime behavior
In case when left part of assignment expression has "side effects", expression like
`data[i++] += 1` is translated to something like `data[i++] = data[i++] + 1` and can cause weird runtime errors.

### Steps To Reproduce

https://deck.net/8fcea6a0ad63ef4caeeda143580f3feb

```c#
public class Program
{
    public static void Main()
    {
        int[] data = {10,20,30};
        int i = 0;
        data[i++] += 1;
        
        Console.WriteLine("data[0] = " + data[0]);
        Console.WriteLine("data[1] = " + data[1]);
        Console.WriteLine("i = " + i);
    }
}
```

### Expected Result

```js
data[0] = 11
data[1] = 20
i = 1
```

### Actual Result

```js
data[0] = 21
data[1] = 20
i = 2
```

vladsch:
Fixes #3063


Bridge/Resources/.generated/system/collections/bitArray.js
```diff
@@ -152,183 +153,184 @@
...
-            $ctor5: function (values) {
-                this.$initialize();
-                if (values == null) {
-                    throw new System.ArgumentNullException("values");
-                }
-                // this value is chosen to prevent overflow when computing m_length
-                if (values.length > 67108863) {
-                    throw new System.ArgumentException(System.String.format("The input array length must not exceed Int32.MaxValue / {0}. Otherwise BitArray.Length would exceed Int32.MaxValue.", Bridge.box(System.Collections.BitArray.BitsPerInt32, System.Int32)), "values");
-                }
+        },
```
LeonidVeriga: @vladsch, there is incorrect offset (indent level), probably due to incorrect formatting with  `var $t` in line https://github.com/bridgedotnet/Bridge/pull/3080/files#diff-8e80b6a2afcc37c7ea6da8e15d3d86e4R145

PR #3096, opened 2017-08-29
[3095] Parallel SharpSixRewriter

### LINKED ISSUE #3095: Improve compilation time
Hi. After some performance measurements, I have noticed that half of compilation process is spent into SharpSixRewriter.Rewrite. It very looks like a case for parallel execution. It's takes immutable SyntaxTree and return the rewritten code. 

So running it parallel gave this results 
**Rewrite: 00:01:03.7732088(CPU Time) / 00:00:28.6653381(Actual Time)**
Total: 00:02:1.6932611(CPU Time) / 00:01:38.3699835(Actual Time)

AndreyZM:
Fixes #3095.

### DIFF
--- a/Compiler/Translator/Translator/Translator.InspectAssembly.cs
+++ b/Compiler/Translator/Translator/Translator.InspectAssembly.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System;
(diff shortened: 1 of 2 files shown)


LeonidVeriga:
@AndreyZM, thank you for the changes provided! We had to create a separate PR #3118 as some changes required and this PR is created on your master branch.

PR #3102, opened 2017-08-31
[2976] `combineScripts: true` on a project that is referenced results in duplication of scripts and fails to load

### LINKED ISSUE #2976: `combineScripts: true` on a project that is referenced results in duplication of scripts and fails to load
Given 3 Bridge projects, `A`, `B` and `C`; `A` references both `B`+`C` and `B` references `C`. `combineScripts: true` is set for all packages.

The build output of `A` gets the JavaScript from `C` twice, once from its own reference and once from the combined scripts in `B`.

This results in an error when loading project A in the browser:

> Class `C.TestType` is already defined

I guess this is because C's JS is being combined  into B's before it then gets embedded in the assembly, and then is pulled out by A. This means currently it's not possible to use `combineScripts` on a project that gets referenced by another if there's a project referenced by both (this is really common with a library like Bridge.React).

It could be argued that we shouldn't use `combineScripts` in non-leaf node projects, however we need it in our production apps and we have test projects that need to reference those production apps.

I think the issue is related to combineScripts being used to create the JS on disk and that then embedded in the DLLs. I wonder if it would be possible to embed "clean" (non-combined) JS into the DLL regardless of the combineScripts setting? (It seems strange to be embedding things like Bridge.js and other dependencies JS inside a DLL).

LeonidVeriga:
Fixes #2976


LeonidVeriga:
@DanTup, this is a PR with #2976 implementation. It would be greate if you could check it out and leave your feedback.

DanTup:
@calcock Do you still have a simple repro for this you can test while I'm OOO?

Issue #1815, opened 2016-09-13
Support for FileReader


Porl91:
https://developer.mozilla.org/en/docs/Web/API/FileReader

DanTup:
Also should include [File](https://developer.mozilla.org/en-US/docs/Web/API/File) and [FileList](https://developer.mozilla.org/en-US/docs/Web/API/FileList) (and if it isn't already done, [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData)).

These are to aid building a drag/drop file upload component.

geoffreymcgill:
The **FileReader** class is part of a _Working Draft_ spec and as a general rule we wouldn't add this API directly in Bridge. 

Probably better as **Bridge.Html5.FileReader** NuGet package. 

The following conversation discusses a similar scenario: https://github.com/bridgedotnet/Bridge/issues/1728#issuecomment-245806403

DanTup:
Turns out we don't need this for what we wanted (file upload), it's only for reading the files locally in the browser. So there's no demand for this from us for now! (@Porl91 is a colleague).

samfromdeath:
Hey @DanTup, They added ContentType Any<string, bool> to AjaxSettings, To support File Upload via Ajax.

michaelcheers:
Try [this](https://pastebin.com/FN7hDert).

PR #3108, opened 2017-09-05
[779] System.IO support

vladsch:
* #779 **System.IO** support
* #1815 Support for **FileReader**
* #3112 Variable from using scope within a loop fails in lambda
* `Script.IsNode`
* `Encoding._hasError`
* `class System.ArraySegment` -> `struct System.ArraySegment`
* **System.Decimal**:
  * `internal static extern byte[] GetBytes(decimal value)`
  * `internal static extern decimal FromBytes(byte[] bytes)`


michaelcheers:
@vladsch, #2789 is also System,IO support.

PR #3193, opened 2017-10-10
[3181] Illegal typescript generated when generating modules

### LINKED ISSUE #3181: Illegal typescript generated when generating modules
Typescript interfaces and vars aren't allowed to have dots '.' in them, however specifying the Module attribute generates such d.ts files - which renders them unusable.

### Steps To Reproduce

https://deck.net/b1857d505d7329994c1e3f11e6e7bb9e

```csharp
[Module(Name="someModule")]
public class Program
{
    public static void Main()
    {
        var msg = "Hello, World!";
        
        Console.WriteLine(msg);
    }
}
```

### Expected Result

```ts

declare module someModule.Demo {
    export interface Program {
    }
        prototype: Program;
        new (): Program;
        Main(): void;
    }
    var Program: ProgramFunc;
}

```

### Actual Result

```ts

declare module Demo {
    export interface someModule.Program {
    }
    export interface someModule.ProgramFunc extends Function {
        prototype: someModule.Program;
        new (): someModule.Program;
        Main(): void;
    }
    var someModule.Program: someModule.ProgramFunc;
}
```

vladsch:
Fixes #3181


fabriciomurta:
image
Reminder: Above issues should be fixed before this can be merged.

PR #3214, opened 2017-10-20
[3207] AutoProperty rules config in bridge.json does not work

### LINKED ISSUE #3207: AutoProperty rules config in bridge.json does not work
I set the "autoProperty" rule in the bridge.json file to "Plain". It did not change the auto-properties to fields as expected. When I added a "Rules" attribute to the AssemblyInfo.cs that sets "AutoProperty" to "Plain", it worked. Am I doing something wrong?

### Steps To Reproduce

bridge.json
```js
{
  "rules": {
    "autoProperty": "Plain"
  }
}
```

### Expected Result

All auto-properties should be compiled to fields.

### Actual Result

All auto-properties remained as properties.

vladsch:
Fixes #3207


fabriciomurta:
Changes from period to comma character as decimal separator to reference files are triggering NUnit tests errors. Probably a globalization issue.

Example of change that makes it break:
image

fabriciomurta:
.gitignore disallowed us from graciously adding *.log files to integration tests' reference data folders. Inclusion of the excluded extension added to .gitignore.

PR #3215, opened 2017-10-20
[3209] Reference declaration files originated from Retyped packages

### LINKED ISSUE #3209: Reference declaration files originated from Retyped packages
A Retyped package holds the declaration files, that were used to generate C# bindings. They can be copied to the output folder when [generateTypeScript](https://github.com/bridgedotnet/Bridge/wiki/global-configuration#generatetypescript) option is set to `true`.

The problem is that the **d.ts** files generated by Bridge for C# code does not reference Retyped **d.ts** files.

### Steps To Reproduce

```csharp
// Requires 'Retyped.knockout' package

namespace Program
{
    using static Retyped.knockout;

    public class App
    {
        public void Foo(KnockoutComponents arg)
        {
        }
    }
}
```

### Expected Result (Program.d.ts)

```ts
/// <reference path="./knockout/index.d.ts" />

declare module Program {
    export interface App {
        Foo(arg: KnockoutComponents): void;
    }
    export interface AppFunc extends Function {
        prototype: App;
        new (): App;
    }
    var App: AppFunc;
}
```

### Actual Result (Program.d.ts)

```ts

declare module Program {
    export interface App {
        Foo(arg: KnockoutComponents): void;
    }
    export interface AppFunc extends Function {
        prototype: App;
        new (): App;
    }
    var App: AppFunc;
}
```

vladsch:
Fixes #3209

### DIFF
--- a/Compiler/Translator/Emitter/Emitter.Output.cs
+++ b/Compiler/Translator/Emitter/Emitter.Output.cs
@@ -232,11 +232,6 @@ namespace Bridge.Translator
 
                     level = this.AddDependencies(level, output, tmp, endOutput);
                 }
-                else
-                {
-                    string newLine = Bridge.Contract.XmlToJSConstants.DEFAULT_LINE_SEPARATOR.ToString();
-                    tmp.Insert(0, @"/// <reference path=""./bridge.d.ts"" />" + newLine + newLine);
-                }
 
                 var code = output.NonModuletOutput.ToString();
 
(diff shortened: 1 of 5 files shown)


fabriciomurta:
The first commit ignores line ending and always sends CRLF, resulting in mixed line endings in output files for specific Bridge settings.

fabriciomurta:
The branch was rebased and fixed, and already merged in.

PR #3233, opened 2017-11-03
[3155] Default logging of Bridge compiler is super-verbose

### LINKED ISSUE #3155: Default logging of Bridge compiler is super-verbose
We have a large number of Bridge projects and currently our Output window in VS (as well as logged output on the build server) is incredibly verbose.

I know this can be controlled in bridge.json, but it feels like it might be a bit noisey by default. The normal MSBuild output is fairly quiet, so it seems like Bridge should be similar by default (or ideally, use the switch already passed to msbuild).

geoffreymcgill:
Fixes #3155

New successful rebuild logging has been cleaned up and reduced to the following. 

```
------ Rebuild All started: Project: Demo1, Configuration: Debug Any CPU ------
  Demo1 -> c:\users\user\Demo1\bin\Debug\Demo1.dll
  Bridge started
      To enable further logging use config setting "logging" in bridge.json.
      https://github.com/bridgedotnet/Bridge/wiki/global-configuration#logging
  Bridge done
========== Rebuild All: 1 succeeded, 0 failed, 0 skipped ==========
````

### DIFF
--- a/.build/files/Bridge.Min.targets
+++ b/.build/files/Bridge.Min.targets
@@ -17,7 +17,7 @@
   </PropertyGroup>
 
   <Target Name="BridgeNetCompilerTask">
-    <Message Text="BridgeNetCompilerTask started: Project: $(MSBuildProjectName), Configuration: $(Configuration) $(Platform) OutputPath: $(OutputPath)" Importance="high" />
+    <Message Text="Bridge started: OutputPath: $(OutputPath)" Importance="high" />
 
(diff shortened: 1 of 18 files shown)


fabriciomurta:
The first commit in this branch requires also a change in bridge's CLI at [Program.cs:68](https://github.com/bridgedotnet/CLI/blob/a74f4992e3f1d528bae7338d88165017dd528fc9/Bridge/Program.cs#L68).

PR #3246, opened 2017-11-05
[3244] Lists aren't IEnumerable of implementing generic types

### LINKED ISSUE #3244: Lists aren't IEnumerable of implementing generic types
Lists aren't of implementing types.

### Steps To Reproduce

https://dev.deck.net/99b8c62328f4364c49a709f5accf4d2d/third
https://dotnetfiddle.net/P2oXGS

```csharp
using System;
using System.Collections.Generic;

public class A{}

public class B : A{}

public static class Program
{
	public static void Main(string[] args)
	{
		object listB = new List<B> {
			new B(),
			new B()
		};
		
		Assert.True(listB is IEnumerable<object>);
		Assert.True(listB is IEnumerable<A>);
	}
}

public static class Assert
{
	public static void True (bool value)
	{
		Console.WriteLine(value ? "Passed" : "Failed");
	}
}
```

vladsch:
Fixes #3244


michaelcheers:
@vladsch, thanks for the quick fix.

PR #3262, opened 2017-11-09
[3208] generated typescript for modules is not a module

### LINKED ISSUE #3208: generated typescript for modules is not a module
The **d.ts** file doesn't generate a valid module since it does not contain a top-level export.

this makes it impossible to import a bridge.net generated library from typescript code.

My goal is to generate an npm package from the output of Bridge.NET project that has a Module attribute (on the entire assembly), and then install the package on my typescript project and use typescript import statement to use that library.

Right now getting the following error from ts:

```
[ts] File 'c:/projects/testBridge/node_modules/test-bridge/test-bridge.d.ts' is not a module.
```

### Steps To Reproduce

https://deck.net/cd409cfda8617388a3a9f1e26f9beae3

```csharp
[Module("testBridge")]
public class Program
{
    public static void Main()
    {
        var msg = "Hello, World!";
        
        Console.WriteLine(msg);
    }
}
```

### Expected Result

```ts

declare module testBridge.Demo {
    export interface Program {
    }
        prototype: Program;
        new (): Program;
        Main(): void;
    }
    var Program: ProgramFunc;
}
export = testBridge.Demo;
```

### Actual Result

```ts

declare module testBridge.Demo {
    export interface Program {
    }
        prototype: Program;
        new (): Program;
        Main(): void;
    }
    var Program: ProgramFunc;
}
```

vladsch:
Fixes #3208


AndreyChechel:
Another issue for AMD module:

### C# sample
the same as above, except module type:
```cs
[Module(ModuleType.AMD, "TestModule")]
public static class TestModuleScope 
{
   // ...
}
```

### TypeScript declarations:
```ts
declare namespace Demo {
    export interface App {
    }
    export interface AppFunc extends Function {
        prototype: App;
        new (): App;
        Main(): void;
    }
    var App: AppFunc;
} // <----- ERROR, that brace was missing

declare module "TestModule" {
    // ...
}
```

AndreyChechel:
Looks good except a minor defect for AMD modules. AMD module name shouldn't be included into a type qualified name:
```ts
declare module "TestModule" {
    namespace Demo {
        export interface TestModuleScope {
        }
        export interface TestModuleScopeFunc extends Function {
            prototype: TestModuleScope;

            // Actual:
            Welcomer: TestModule.Demo.TestModuleScope.WelcomerFunc;
            // Expected:
            Welcomer: Demo.TestModuleScope.WelcomerFunc;

            new (): TestModuleScope;
        }
        var TestModuleScope: TestModuleScopeFunc;
        module TestModuleScope {
            export interface Welcomer {
                Hello(name: string): void;
            }
            export interface WelcomerFunc extends Function {
                prototype: Welcomer;
                new (): Welcomer;
            }
        }
    }
}
```

PR #3263, opened 2017-11-09
[3258] Missing unary operator with Template implicit operator

### LINKED ISSUE #3258: Missing unary operator with Template implicit operator
### Steps To Reproduce

https://deck.net/148dae798f29412ec33dc84293d0869c

```csharp

public class O
{
    [Bridge.Template("{d}")]
    static extern public implicit operator O(double d);
}

public class Program
{
    public static void Main()
    {
        double i = 0;
        O a = -(i);//missing operator here
        O b = (-(i));//but not here
    }
    
     public static double mock(){return 0.0f;}
}
```

### Expected Result

```js
/**
 * @compiler Bridge.NET 16.4.1
 */
    "use strict";

    Bridge.define("Demo.O");

        main: function Main () {
            var i = 0;
            var a = -(i); //missing operator here
            var b = (-(i)); //but not here
        },
        statics: {
            methods: {
                mock: function () {
                    return 0.0;
                }
            }
        }
    });
});

```

### Actual Result

```js
/**
 * @compiler Bridge.NET 16.4.1
 */
    "use strict";

    Bridge.define("Demo.O");

        main: function Main () {
            var i = 0;
            var a = (i); //missing operator here
            var b = (-(i)); //but not here
        },
        statics: {
            methods: {
                mock: function () {
                    return 0.0;
                }
            }
        }
    });
});

```

vladsch:
Fixes #3258

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -22684,7 +22684,7 @@ Bridge.Class.addExtend(System.String, [System.IComparable$1(System.String), Syst
 Bridge.define("System.Text.RegularExpressions.RegexParser", {
     statics: {
         _Q: 5, // quantifier
-        _S: 4, // ordinary stoppper
+        _S: 4, // ordinary stopper
         _Z: 3, // ScanBlank stopper
         _X: 2, // whitespace
         _E: 1, // should be escaped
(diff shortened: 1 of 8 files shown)


fabriciomurta:
The same test case, if turned into integers, results the next code in the following Javascript output. The negative result is kept nevertheless (as the bitwise or does nothing to already integer numbers).

```
int j = 1;
O x = -j;
O y = -(j);
O z = (-(j));
```

Outputs this:
```
var j = 1;
var x = (-j) | 0;
var y = (-(j)) | 0;
var z = (((-(j)) | 0));
```

This probably relates to keeping an integer an integer under bridge [by this stackoverflow discussion](https://stackoverflow.com/questions/7487977/using-bitwise-or-0-to-floor-a-number).

PR #3277, opened 2017-11-16
[3276] More specific types for generated .d.ts

### LINKED ISSUE #3276: More specific types for generated .d.ts
Given the following sample, the generated .d.ts is specifying a `number` type, but it should be a `System.Decimal` type.

There may be other scenarios, especially where integer types are used, that should be updated to output specific types instead of `number`. A general review is required.

### Steps To Reproduce

https://deck.net/88ff64cb03e8ec29ba6475d9e1264a67

```csharp
public class Program
{
    public static decimal Demo(decimal a, decimal b)
    {
        return a * b;
    }
}
```

### Expected Result

```js
Demo(a: System.Decimal, b: System.Decimal): System.Decimal;
```

### Actual Result

```js
Demo(a: number, b: number): number;
```

## See Also

* #3267

vladsch:
Fixes #3276


fabriciomurta:
Last changes looking good. Comment blocks' output is as expected!

PR #3295, opened 2017-11-30
[3292] Updates necessary to fix issue

### LINKED ISSUE #3292: Implicit operator not called with interface implementations
### Steps To Reproduce

 https://deck.net/82c05d267049e5ead57d293cfca06817

```csharp
public class Program
{
    public static void Main()
    {
        // This should generate JavaScript that calls the implicit Optional<T> operator but it doesn't -
        // the "new Demo.GenericError()" is passed directly into the UpdateAction constructor
        // Note: If the UpdateAction constructor takes an Optional of GenericError instead of an
        //       Optional of IApiErrorDetails then the implicit cast call is correctly emitted
        var shouldUseImplicitCast = new UpdateAction(new GenericError());
    }
}

public sealed class UpdateAction
{
    public UpdateAction(Optional<IApiErrorDetails> errorIfFailed) { }
}

public interface IApiErrorDetails { }

public sealed class GenericError : IApiErrorDetails { }

public struct Optional<T>
{
    public Optional(T value)
    {
        Value = value;
    }

    public T Value { get; }

    public static implicit operator Optional<T>(T source)
    {
        return new Optional<T>(source);
    }
}
```

### Expected Result

```js
var shouldUseImplicitCast = new Demo.UpdateAction(Demo.Optional$1(Demo.IApiErrorDetails).op_Implicit(new Demo.GenericError()));
```

### Actual Result

```js
var shouldUseImplicitCast = new Demo.UpdateAction(new Demo.GenericError());
```

### See Also
https://forums.bridge.net/forum/bridge-net-pro/bugs/5019-implicit-operator-not-called-with-interface-implementations

fabriciomurta:
Fixes #3292 .

- Update bridge references to NRefactory 5.5.9
- Add unit test for the specific issue (pending)

### DIFF
--- a/Compiler/Contract/Bridge.Contract.csproj
+++ b/Compiler/Contract/Bridge.Contract.csproj
@@ -39,16 +39,13 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="ICSharpCode.NRefactory, Version=5.0.0.0, Culture=neutral, PublicKeyToken=d4bfe873e7598c49, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Bridge.NRefactory.5.5.8\lib\net40\ICSharpCode.NRefactory.dll</HintPath>
-      <Private>True</Private>
+      <HintPath>..\..\packages\Bridge.NRefactory.5.5.9\lib\net40\ICSharpCode.NRefactory.dll</HintPath>
     </Reference>
     <Reference Include="ICSharpCode.NRefactory.Cecil, Version=5.0.0.0, Culture=neutral, PublicKeyToken=d4bfe873e7598c49, processorArchitecture=MSIL">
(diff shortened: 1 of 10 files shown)


fabriciomurta:
Tests are not going to pass until Bridge.NRefactory 5.5.9 is published to NuGet.

There should be a way to trigger an exception because of the different/unexpected code output: https://deck.net/02d5c17ca01019092947aaacaca8b421

fabriciomurta:
@vladsch suggested a nice test case scenario with https://deck.net/6c94f61e9e7fe78f26e23d5d1d54db68

Once built as an unit test, before linking to NRefactory 5.5.9:
image

fabriciomurta:
NRefactory 5.5.9 has been pushed to NuGet so this should work now.

PR #3299, opened 2017-12-03
Run translator in Bridge task using separate app domain

vladsch:
Bridge task uses app domain now to run Bridge translator (should help with memory leaking)


fabriciomurta:
This is not working at all. If you build Bridge then use its NuGet package in a fresh project, it will break during build task in an error claiming it can't find "Microsoft.Build.Framework". It does not display detailed messages (should need to implement further logging messages).

The same behavior happens also if the last commit from @AndreyChechel (dfee9e4) is reverted (using just f7b386f).

.build/specs/Bridge.Min.nuspec
```diff
@@ -29,6 +29,7 @@
 
     <!--tools-->
     <file src="../../Compiler/Build/bin/Release/Bridge.Build.dll" target="tools/Bridge.Builder.v16.dll" />
+	<file src="../../Compiler/Build/bin/Release/Bridge.Build.dll" target="tools/Bridge.Build.dll" />
```
fabriciomurta: Is this change supposed to be like this? The same file copied over to two?

vladsch: Yes, it is required because renamed assembly is not found under separate app domain,

PR #3302, opened 2017-12-06
Use `BridgeNetCompilerTask` when running on MSBuild on *nix

kekekeks:
I have no idea what was the problem with running msbuild task with xbuild before, but now it seems to be working just fine with Mono 5.2 and MSBuild 15 (actual MSBuild, not xbuild)


fabriciomurta:
Thanks for the pull request! This looks like a safe change, as it will still retain calling the tool until it ensures `msbuild` is present in the system.

fabriciomurta:
@kekekeks have you actually built bridge on OSX? No matter I use homebrew or mono-develop package, I just get NuGet packages failing to be located by the NuGet tool's restore. Although I can see the respective packages' files in the cache directory (meaning it actually downloaded them).

kekekeks:
You are getting restore errors because of CA certificate issues in Mono and (probably) outdated NuGet version in Bridge repo. Build works like a charm on Ubuntu 16.04. Try importing system certificates with `cert-sync` utility.

BTW, including `nuget.exe` to your repo is deprecated. Bridge also uses an ancient SDK version (14.0) instead of a new one.

kekekeks:
I'd recommend to migrate your entire build to .NET Core SDK, it still has support for `net40` target.

kekekeks:
You can check that you've properly imported certificates by running `mono .nuget/nuget.exe  update -self`

fabriciomurta:
well, that was not SSL certs, but just nuget client version. I'm working in another pull request to make the thing build with mono at once. But for this one not thinking on revamping the whole thing to .NET Core SDK.
Thanks for your suggestions, was jut wondering if it worked "out of the box" for you.

fabriciomurta:
Thanks for the pull request! Looks great! That's nice `msbuild` is now working better unix platforms.

Issue #3259, opened 2017-11-07
Out of Memory Exceptions thrown during compilation


Porl91:
We have about 8 Bridge projects currently in our main solution and recently we've been getting some intermittent `OutOfMemoryException` exceptions thrown when rebuilding. 

I'm not really sure how best to profile this to understand where the memory's going, although the stack traces do seem to be coming from the BridgeCompilerTask.

Here are some of the stack traces we've received recently:

<pre>System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at Mono.Cecil.PE.ByteBuffer.Grow(Int32 desired)
   at Mono.Cecil.PE.ByteBuffer.WriteBytes(Byte[] bytes)
   at Mono.Cecil.MetadataBuilder.AddResources()
   at Mono.Cecil.MetadataBuilder.BuildModule()
   at Mono.Cecil.ModuleWriter.<>c.<BuildMetadata>b__1_0(MetadataBuilder builder, MetadataReader _)
   at Mono.Cecil.ModuleDefinition.Read[TItem,TRet](TItem item, Func`3 read)
   at Mono.Cecil.ModuleWriter.BuildMetadata(ModuleDefinition module, MetadataBuilder metadata)
   at Mono.Cecil.ModuleWriter.WriteModuleTo(ModuleDefinition module, Stream stream, WriterParameters parameters)
   at Mono.Cecil.ModuleDefinition.Write(Stream stream, WriterParameters parameters)
   at Mono.Cecil.ModuleDefinition.Write(String fileName, WriterParameters parameters)
   at Bridge.Translator.Translator.EmbeddResources(List`1 resourcesToEmbed)
   at Bridge.Translator.Translator.InjectResources(String outputPath, String projectPath)
   at Bridge.Translator.TranslatorProcessor.PostProcess()
   at Bridge.Build.BridgeCompilerTask.Execute()</pre>

<pre>System.OutOfMemoryException: Exception of type 'System.OutOfMemoryException' was thrown.
   at Mono.Cecil.PE.ImageReader.ReadSectionData(Section section)
   at Mono.Cecil.PE.ImageReader.ReadSections(UInt16 count)
   at Mono.Cecil.PE.ImageReader.ReadImage()
   at Mono.Cecil.PE.ImageReader.ReadImageFrom(Stream stream)
   at Mono.Cecil.ModuleDefinition.ReadModule(Stream stream, ReaderParameters parameters)
[...]

PR #3305, opened 2017-12-11
[3259] Avoids resolving already resolved references in Translator.

fabriciomurta:
This was a suggestion from andersnm in bridgedotnet/Bridge#3259 to address
memory leaks during Bridge builds.

It consists in checking if a reference being loaded is already resolved and,
if so, uses the already resolved instance instead of resolving it again (thus
allocating even more memory)

This mostly affects projects with several references and cross-references
memory allocation by bridge during build time. Anyway, unnecessary
references resolving is replaced by a string search.


fabriciomurta:
The additional code suggested by @andersnm does not break anything as far as current bridge tests are concerned, but the fact it does not resolve again no matter the version may pose a problem when referenced projects references to other versions of the same assembly referenced in the root project.

fabriciomurta:
#3314 was merged, this one shouldn't.

Issue #3307, opened 2017-12-12
DateTime sorting fails after modifying (via AddMinutes(), etc.)


yctri:
IOrderedEnumerable can't compare DateTimes after they've been modified with AddMinutes(), etc. It attempts to compare ticks that have been removed from DateTime instances (as a result of being modified by their adjustment methods).

### Steps To Reproduce

https://deck.net/2a202202c30951cc29fb109379f42ee0

```csharp
public class App
{
    public static void Main()
    {
        List<DateTime> times = new List<DateTime>();
        
        DateTime dt1 = DateTime.UtcNow;
        times.Add(dt1);
        DateTime dt2 = dt1.AddMinutes(-10);
        times.Add(dt2);
        
        times = times.OrderBy(dt => dt).ToList();
        
        Console.WriteLine(times[0]);
        Console.WriteLine(times[1]);
    }
}
```

### Expected Result

```js
12/12/2017 00:05:45
12/12/2017 00:15:45
```

### Actual Result

```js
System.Exception: TypeError: Cannot read property 'low' of null
    at o (https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:119820)
    at i.n.compare (https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:122194)
    at System.Int64.compareTo (https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:128325)
    at Object.compare (https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:17168)
    at Object.compare (https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:16772)
    at ctor.<anonymous> (https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:235941)
    at v.compare (https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:361451)
    at https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:360913
    at Array.sort (native)
    at https://deck.net/resources/js/bridge/bridge.min.js?16.5.0:7:360885
```

yctri:
I think this might be because the OrderBy() clause doesn't call the DateTimes' comparison functions(?) It just does an integer comparison on DateTime.Ticks. The latter fails because of another issue: modifying a DateTime (such as with AddMinutes()) removes its Ticks member in javascript.

fabriciomurta:
Fixed and live in Bridge 16.6.1! Thanks!

Issue #3306, opened 2017-12-11
Adjusting a DateTime and then calling 'new DateTime(ticks, kind)' returns wrong time


yctri:
When adding time to an existing UTC DateTime and then using the DateTime(ticks, kind) constructor to reconstruct the same DateTime, the new instance no longer appears to reflect UTC time.

### Steps To Reproduce

https://deck.net/3f7aea2477c59ccd77b658b0016f9752

```csharp
public class App
{
    public static void Main()
    {
        DateTime dt = DateTime.UtcNow;
        
        Console.WriteLine(dt);    // Outputs 12/11/2017 22:57:38
        
        dt = dt.AddMinutes(1);
        dt = new DateTime(dt.Ticks, dt.Kind);
        
        Console.WriteLine(dt);    // Outputs 12/11/2017 14:58:38 
    }
}
```

### Expected Result

12/11/2017 22:57:38
12/11/2017 22:58:38

### Actual Result

12/11/2017 22:57:38
12/11/2017 14:58:38

fabriciomurta:
The issue got closed by an issue's side effect message and I missed posting the feedback here! Glad I added the "in progress" tag to the issue. :)

Anyways, it is already live in latest published Bridge, 16.6.1! Thanks for the fix!

PR #3325, opened 2017-12-21
Fixed DateTime problems in Date.js

yctri:
- Fixes https://github.com/bridgedotnet/Bridge/issues/3306
- Fixes https://github.com/bridgedotnet/Bridge/issues/3307
- Fixes bad DateTime when parsing UTC (Z) times.
- Fixes Date extraction not taking DateTimeKind into account.
- Mutating DateTime now sets missing Ticks member.

(First time doing a pull request here, so apologies in advance if I missed anything)


geoffreymcgill:
Hi @yctri – Thanks for submitting this pull-request. 

Unfortunately, we cannot accept this pr into the 16.6.0 release, but our plan is to review immediately after the 16.6.0 release, then release a 16.6.1 with this pr.

fabriciomurta:
You should be able to build and run the tests if you build **Bridge.Dev.sln** and run (view in browser) the **Tests/Runner** project.

yctri:
Good to know, thanks. I won't be able to look into the failing unit tests for a while, so do feel free to take a stab at this if the fix is obvious to you guys...

Thanks

fabriciomurta:
Hello and thanks for your feedback! We'll try on our own to see about the new issues when we get a chance.

But if you beat us to it, you're more than welcome to draw out the negative side effects of the changes. Besides, being able to run the tests on your own would greatly help you further contribute in the future.

yctri:
Hi @fabriciomurta, I managed to resolve all of the unit test failures with the latest commit. 

Presumably during Date construction, JS converts the caller's specified time to UTC time internally before calculating its milliseconds since Unix epoch, whereas .NET's DateTime does not (since the latter relies on DateTimeKind instead). Therefore, getTicks() now checks Kind to determine whether the timezone offset was applied internally before offsetting ticks. 

Thanks for pointing out the unit tests!

fabriciomurta:
Well, seems github is not always as smart as we expect!.. To preserve your commits as they are, I've rebased your changes to our 16.6.1 branch (we already moved `dev` into 16.7.0 with a new feature, but we really want to make a 16.6.1 with your fix, so we branched off with also other minor fixes).
Anyway, your fix has been merged and we're going to release it.
Although the pull request will read as "closed" it has been effectively merged, it is just not always github can detect a merged pull request if they didn't follow the "vanilla protocol". ;)

But see your commits here in history: https://github.com/bridgedotnet/Bridge/commits/version16.6.1

PR #3336, opened 2017-12-29
Handle modules defined in bridge.json the same way as attribute modules

### LINKED ISSUE #3335: Invalid module handling if module is defined in bridge.json
It appears that in case you define your module in the bridge.json the output is different than if you define the module via ModuleAttribute. The types within the assembly are then not implicitely in the module and therefore lack of the required $scope. 

I detected this bug during one of my current Bridge.net code cleanups. I "broke" the existing IntegrationTest 7 by "correctly" handling the module definitions. This integration test defines a module in the bridge.json but the types are not included in the module causing a different output. 

### Steps To Reproduce

Deck: Not possible because assembly level attributes and bridge.json cannot be specified on deck. 

#### Module in Assembly

```csharp
[assembly:Bridge.Module("MyModule")]
namespace TestProject
{
    public class N3334
    {
    }
}
```
#### Module in Json

```json
{
  "output": "Bridge\\output",
  "cleanOutputFolderBeforeBuildPattern": "*.*",
  "outputBy": "namespace",
  "module": {
    "name": "MyModule"
  },
  "outputFormatting": "Formatted",
  "generateDocumentation": "Full"
}
```

```csharp
namespace TestProject.Issues 
{
    public class N3334
    {
    }
}
```

### Expected Result

```js
/**
 * @compiler Bridge.NET 16.6.1
 */
Bridge.assembly("TestProject", function ($asm, globals) {
    "use strict";

    define("MyModule", function () {
        var MyModule = { };
        /** @namespace TestProject */

        /**
         * @public
         * @class MyModule.TestProject.N3334
         */
        Bridge.define("TestProject.N3334", {
            $scope: MyModule,
            statics: {
                methods: {
                    $metadata : function () { return {"att":1048577,"a":2,"m":[{"a":2,"isSynthetic":true,"n":".ctor","t":1,"sn":"ctor"}]}; }
                }
            }
        });
        return MyModule;
    });

});

```
#### Module in Assembly

```js
/**
 * @compiler Bridge.NET 16.6.1
 */
Bridge.assembly("TestProject", function ($asm, globals) {
    "use strict";

    define("MyModule", function () {
        var MyModule = { };
        /** @namespace TestProject */

        /**
         * @public
         * @class MyModule.TestProject.N3334
         */
        Bridge.define("TestProject.N3334", {
            $scope: MyModule,
            statics: {
                methods: {
                    $metadata : function () { return {"att":1048577,"a":2,"m":[{"a":2,"isSynthetic":true,"n":".ctor","t":1,"sn":"ctor"}]}; }
                }
            }
        });
        return MyModule;
    });

});
```

#### Module in JSON

```js
/**
 * @compiler Bridge.NET 16.6.1
 */
Bridge.assembly("TestProject", function ($asm, globals) {
    "use strict";

    define("MyModule", function () {
        var MyModule = { };
    /** @namespace TestProject */

    /**
     * @public
     * @class TestProject.N3334
     */
    Bridge.define("TestProject.N3334");
        return MyModule;
    });

});
```

### Hints for fix

The `Validator.CheckModule` does not consider the `AssemblyInfo.Module`. https://github.com/bridgedotnet/Bridge/blob/9cffa0be5546053ea11f13001f304825bdd1734b/Compiler/Translator/Inspector/Validator.cs#L727

This in a result causes the module to be missing in the `EmitBlock.DoEmit`:

https://github.com/bridgedotnet/Bridge/blob/9cffa0be5546053ea11f13001f304825bdd1734b/Compiler/Translator/Emitter/Blocks/EmitBlock.cs#L289

I fear this bug also has an impact on the other `ITypeInfo` properties like `FileName` and `Dependencies`.

Danielku15:
Fixes #3335  .

- Handle modules defined in bridge.json the same way as attribute modules


fabriciomurta:
Hello! Thanks for the pull request! It looks good, but the first commit was just unnecessary, it just added some assert() lines because you built in Debug mode.

Wouldn't it be a good idea to have a dedicated Issue3335 integration test for this aspect. Maybe two tiny projects implementing each of the alternatives...

Danielku15:
> It looks good, but the first commit was just unnecessary, it just added some assert() lines because you built in Debug mode.
I was previously working on an older commit of the dev-branch where the asserts/contracts were not included (also not in debug). After I synced again they appeared again, maybe as a result of a bugfix. To ensure that my changes did not have an impact on the output I reverted my changes and first made a commit with the outputs updated. 

> Wouldn't it be a good idea to have a dedicated Issue3335 integration test for this aspect. 
I added 2 dedicated integration tests that ensure the output.

Issue #3333, opened 2017-12-26
Contribution Guidelines: Code Style and Architectural Cleanup


Danielku15:
I would like to know if there are some sort of official (basic) coding guidelines for code contribution to Bride.net. My Visual Studio (and Resharper) are crying about many code style things which I'd like to get clean. They mainly complain about: 
- The usage of unneeeded `this.` qualifiers which seems to be common. 
- Potential usage of new C#7 features. 

Also I'd like to know if it is safe to cleanup the Bridge codebase. e.g. 
- Remove I detect dead code (even if it is part of the contracts IEmitter, IValidator etc.) 
- Move code to more common places (e.g. use extension methods instead of having them part of the contracts). 

A little bit background to my question: 

I am currenly working on some bigger rework of the Bridge.net compiler. The rework is mainly driven by the issues #394 and #2002. I started implementing the fluent API configuration and as part of this I needed to get rid of Mono.Cecil as part of the compilation process. I am making very good progress but fear a bit that my changes are too big to be accepted as such.

geoffreymcgill:
Hi. I added our [C# Coding Style Guide](https://github.com/bridgedotnet/Bridge/wiki/C%23-Coding-Style-Guide) to the wiki.

The use of `this.` is not required from the compiler perspective, but preferred as it provides an obvious declaration on whether a member is `instance` or `static`.

Thanks for the contribution. We're excited to see your modifications.

PR #3337, opened 2017-12-29
Compile-Time injection of Attributes

### LINKED ISSUE #394: C# Class Library configuration through external Bridge.NET fluent config project
http://forums.bridge.net/forum/community/help/506-code-configuration-specification-instead-additional-to-the-attributes

Assigning to @vladsch, but this is something that several of us will have to discuss and be involved in.

### LINKED ISSUE #2002: Too many dependencies on Mono.Cecil
This is not directly a bug, but causes some troubles on another topic I am currently working on (ExternalAnnotations #394). 

A lot of parts in the code are based on the Mono.Cecil type system (e.g. Mono.Cecil.TypeDefinition) which is loaded from the DLL after compilation of the project (Translator.InspectAssembly.cs). The other parts are relying on the NRefactory type system (e.g. ICSharpCode.NRefactory.TypeSystem.ITypeDefinition). 

The concrete problem I am facing with #394 is that I am dynamically rewriting some syntax trees and adding attributes to code elements. Since a lot of parts of the code rely on the information from the DLL, of course the dynamically added attributes are not considered. 

In my opinion the correct approach should be to use the NRefactory type system everywhere. Are there particular reasons why the output assembly is loaded separately? I would like to clean up the code base from all Mono.Cecil usages and replace them with the NRefactory parts but before doing so,  I would like to hear the decision behind using partly Mono.Cecil.

Danielku15:
Fixes #394, Fixes #2002 .

- Centralized all attribute logic to a common `AttributeRepository`
- Eliminated Mono.Cecil on the compilation process (replaced by NRefactory.TypeSystem)
- Added CompilerExtension feature which allows registering attributes during compile-time through an API. 
- Cleanup of dead code. 

Feature description:

With this change developers can inject attributes during compile-time. This allows developers to separate their attribute/output configuration from the actual code base which might be shared with non-bridge platforms. The attribute registration works via an `ICompilerContext` interfaces. The so called compiler extensions must be part of the compilation process as the registration happens via a custom code interpreter with restricted access to actual C# code. Reason behind this interpreter is that an actual code execution would fail in 90% due to missing implementations in the base BCL and also the JS focused C# code. It can be more seen as C# based configuration file. 

As part of this rework also the Mono.Cecil dependencies are almost eliminated beside actual Mono.Cecil tasks (e.g. resource handling). 

## See Also

* #3333


geoffreymcgill:
Hi @Danielku15 – Could you provide a sample demonstrating with features from this new PR, how developers can inject attributes during compile-time?

Danielku15:
The PR contains an integration test demonstrating the feature: 
https://github.com/Danielku15/Bridge/blob/84953b2f0554d83b94339e63bfe47bd551350b6b/Compiler/TranslatorTests/TestProjects/23/Test.cs

geoffreymcgill:
Hi @Danielku15 – This pr looks amazing. We have a backlog of pr's waiting for review and testing within `v17`, and once pr #3341 is finished too, the team will get started on this pull-request.

PR #3339, opened 2017-12-31
Improved performance by caching typesystem and improved Roslyn rewriter

Danielku15:
During my current development on Bridge I got a bit annoyed that the compilation of the Bridge assembly takes always that long. I digged a bit into the bottlenecks and tried to improve them. 

- [Bridge] Caching of TypeSystem in MemberResolver. 
- [Roslyn] Resolve ITypeSymbol only once on TypeDeclarationSyntax level.
- [Roslyn] Pre-Check if any using statement defines an alias or static include, if yes skip rewriting. 

I am not 100% sure if it is safe to skip the complete `VisitIdentifierName` and `VisitGenericName` in case of no aliases and static includes. The code is a bit hard to understand without any comments. Maybe @vladsch can have a closer look. All unit tests are green. 

Here a before-after comparison of some test runs on my dev machine compiling the bridge assembly:

Compilation Phase | Before | After
---------------------|---------|------
Init | 358ms | 312ms
Building Syntax Tree | 29189ms | 10507ms
Type Inspection | 1527ms | 341ms
Emitting | 3146ms | 3044ms
Output | 2182ms | 1932ms
**Total** | **41016ms** | **19063ms**


geoffreymcgill:
Hi @Danielku15 – Thanks for this pull request. Looks awesome. 

We are reviewing the PR and looking into whether we can include this in the upcoming Bridge 16.7.0 release.

fabriciomurta:
Just wondering how would best limit this code to run under 1 task (for debugging, for example). Else, debugging stuff may become an issue.

fabriciomurta:
You should enable us to push commits to this pr before we can accept it.

Danielku15:
> Just wondering how would best limit this code to run under 1 task (for debugging, for example). Else, debugging stuff may become an issue.

For this we would need to change [Translator.InspectAssembly.cs line 320](https://github.com/bridgedotnet/Bridge/pull/3339/files#diff-d6df2d9fa7d4c54b507060ade848e648R320). It currently paralellizes all rewrites. Potentially a `Parallel.For` could deliver the same result with configurable number threads: 

```csharp
var parallelizmDegree = -1 /*unlimited*/;
#if DEBUG
if (Debugger.IsAttached)
{
	parallelizmDegree = 1;
}
#endif
Parallel.For(0, this.SourceFiles.Count,
	new ParallelOptions
	{
		MaxDegreeOfParallelism  = parallelizmDegree
	},
	index =>
	{
		var fileName = this.SourceFiles[index];
		this.Log.Trace("Source file " + (fileName ?? string.Empty) + " ...");
		//...
		this.Log.Trace("Source file " + (fileName ?? string.Empty) + " done");
	}
);
```

fabriciomurta:
Hello! Thanks for opening the perms. Pushed the updates and squash-merged the changes.

I usually prefer to rebase-merge to keep your original commits as they were, but as we had some commits with changes that needed to be reverted, that would pollute history with stuff that should never be there. So then I just squashed the changes so all reverted output gets "squashed out" in the process.

Thank you very much for the thoughtful improvement!

Oh, about your last message, I talked to @vladsch, and it didn't look like a problem for debugging with the implementation you did, after all. But your suggestion is very close to what I was thinking would be best. If you want to provide it as a new pull request, we'll gladly merge it in!

By the way, when you commit those output/js files, ensure you built bridge in `Release` configuration. Else they would just pollute the history with changes that shouldn't make it there.

This change is going to make it to Bridge 16.7.0!

Issue #3357, opened 2018-01-10
Bridge considers two equiv Uris different (but .NET considers them the same)


DanTup:
```csharp
Console.WriteLine(new Uri("https://deck.net") == new Uri("https://deck.net"));
```

Outputs True in .NET, but False in Bridge.

geoffreymcgill:
Issue has been fixed and we're in the process of final review and testing.

DanTup:
@geoffreymcgill @vladsch Worth nothing that `.absoluteUri` in Bridge doesn't currently normalise casing of the host like .NET does, so I think the PR doesn't handle this:

```cs
Console.WriteLine(new Uri("https://GOOGLE.com/test").AbsoluteUri == new Uri("https://google.com/test").AbsoluteUri);
```

In .NET, this is true. In Bridge, currently it's false (and I don't think the PR fixes that). I had a quick look at the [Uri code](https://referencesource.microsoft.com/#System/net/System/URI.cs,991565dd25f47c90) and it looks somewhat complicated :-(

fabriciomurta:
Nice point, @DanTup!

By the way, any reason you're using `.AbsoluteUri`? I don't seem to see any change in results with or without it.
https://dotnetfiddle.net/USgxce

fabriciomurta:
Actually... Current System.Uri is but a simplified stub and should be actually developed. In fact, issue #2185 is tracking its development.

So probably this casing (and a lot other stuff) should go to that issue.

DanTup:
> By the way, any reason you're using `.AbsoluteUri`?

I'm not, I just looked at the PR to fix this and saw the fix was to use AbsoluteUri for equality check and the idea of case sensitivity popped into my head so I thought I'd test it.

Though until this is released, we'd actually have to use AbsoluteUri as the equality on Uri isn't there yet?

fabriciomurta:
Hello again, @DanTup!

We're going to merge in the fix applied which gives a bare minimum functionality to comparisons using System.Uri. But as far as the full functionality as well as more partial functionality related to the class, it will all be concentrated within the #2185 issue. So we will be closing this one to keep all in a same place.

Don't worry about the missing tests, we added comments to the main issue with a set of tests that ensures those checks will not go unnoticed when it is implemented.

DanTup:
Sounds reasonable to me; the original PR still fixes the original issue we had (we're not likely to end up with different cases and then expect equality, since the source of any uris we expect to be equal will be the same) 👍

PR #3364, opened 2018-01-12
[3301] Uri.ToString() returns "[Object object]"

### LINKED ISSUE #3301: Uri.ToString() returns "[Object object]"
This code behaves differently in .NET/Bridge:

```cs
Console.WriteLine(new Uri("https://deck.net/").ToString());
```

.NET outputs `https://deck.net/` but Bridge outputs `[object Object]`. A workaround is to use `.AbsoluteUri`.

DotNetFiddle: https://dotnetfiddle.net/wLkrNq
Deck: https://deck.net/c401a826ae10dadcd131b6c61bbe71c5

vladsch:
Fixes #3301 and #3357


fabriciomurta:
Just for the record, as far as #3301 is concerned I found several among these tests are failing while checking what becomes of a new Uri() out of an ordinary string and converted back to string:
dotnet fiddle: https://dotnetfiddle.net/0Yy5cu

But this seems to go beyond this pull request's aim, probably a new issue should be filed to elaborate on `System.Uri` logic to mimic .NET's. Currently it is basically a simple string, not doing any conversion/inferring .NET does.

PR #3375, opened 2018-01-18
Linux/OSX building support overhaul

fabriciomurta:
This pull request includes changes that allows Bridge to be built under Linux or OS/X using Mono 5.

The changes also includes a review on NuGet packaging mechanism to allow it also to work under other OS's and other cross-platform fixes necessary.


fabriciomurta:
Bridge fully builds on linux/osx now. But still the tests runner are pending, as well as other repositories. No matter what, this is a good progress towards fully building everything in a platform independent environment.

PR #3377, opened 2018-01-20
Makes EmitBlock output string on double quotes

fabriciomurta:
Since pull request #3353, bridge meta output were getting string values wrapped in single quotes. This set of changes replaces them by double quotes.


fabriciomurta:
The seemingly unrelated changes in `Compiler/TranslatorTests/TestProjects/02/Bridge/reference/TestProject.I2096.js` are because the file is only considered from a line containing `@!!!` onwards, so the tests didn't require updating on other issues that only alter ignored parts of the results file.

PR #3383, opened 2018-01-26
Fix XMLHttpRequestEventTarget's events not all being ProgressEvent

dionrhys:
The events present on `XMLHttpRequestEventTarget` need to be `ProgressEvent` not `Event`, according to the WHATWG spec here: https://xhr.spec.whatwg.org/#events

This change means we don't have to manually cast these event arguments to the correct type.

https://deck.net/ad7792f684b0259078e90aeb4bd2a065

**Expected result:**

```
> {number of bytes downloaded}
```

**Actual result:**

```
<Compilation error\>
```

- Change the type of `XMLHttpRequestEventTarget`'s `OnAbort`, `OnError`, `OnLoad`, `OnLoadEnd`, `OnLoadStart`, and `OnTimeout` events from `Event` to `ProgressEvent`

I haven't tested tests against this pull request because the Bridge.Test project is not open.

### DIFF
--- a/Html5/Ajax/XMLHttpRequestEventTarget.cs
+++ b/Html5/Ajax/XMLHttpRequestEventTarget.cs
@@ -13,25 +13,25 @@ namespace Bridge.Html5
         /// The function to call when a request is aborted.
         /// </summary>
         [Name("onabort")]
-        public Action<Event> OnAbort;
+        public Action<ProgressEvent> OnAbort;
 
         /// <summary>
         /// The function to call when a request encounters an error.
         /// </summary>
         [Name("onerror")]
-        public Action<Event> OnError;
+        public Action<ProgressEvent> OnError;
 
         /// <summary>
         /// The function to call when an HTTP request returns after successfully loading content.
         /// </summary>
         [Name("onload")]
-        public Action<Event> OnLoad;
+        public Action<ProgressEvent> OnLoad;
 
         /// <summary>
         /// A function that gets called when the HTTP request first begins loading data.
         /// </summary>
         [Name("onloadstart")]
-        public Action<Event> OnLoadStart;
+        public Action<ProgressEvent> OnLoadStart;
 
         /// <summary>
         /// A function that is called periodically with information about the progress of the request.
@@ -43,12 +43,12 @@ namespace Bridge.Html5
         /// A function that is called if the event times out; this only happens if a timeout has been previously established by setting the value of the XMLHttpRequest object's timeout attribute.
         /// </summary>
         [Name("ontimeout")]
-        public Action<Event> OnTimeout;
+        public Action<ProgressEvent> OnTimeout;
 
         /// <summary>
         /// A function that is called when the load is completed, even if the request failed.
         /// </summary>
         [Name("onloadend")]
-        public Action<Event> OnLoadEnd;
+        public Action<ProgressEvent> OnLoadEnd;
     }
 }


fabriciomurta:
Isn't it, by saying _"this section is non-normative"_, means that we shouldn't rely on that and that's not necessarily what is implemented on browsers?

fabriciomurta:
For the record, `Console.WriteLine(((ProgressEvent)ev).Load);` on chrome returns a value, suggesting at least chrome supports the type. The issue is just, is this within the "living standard" we should follow?

And here's a MDN definition for the non-normative sections, which should probably be the same within whatwg.
https://developer.mozilla.org/en-US/docs/Glossary/non-normative

dionrhys:
> Isn't it, by saying "this section is non-normative", means that we shouldn't rely on that and that's not necessarily what is implemented on browsers?

By that line of thought, `OnProgress` would also have to just be declared an `Event` instead of `ProgressEvent` because that's also under the "non-normative" section.

Since all of these events are listed in the table as `ProgressEvent`, I think they should all follow the same in Bridge.

It may be nice to provide tests that somehow assert the correct event type is given when these event handlers are called. I'm not sure how this could be done other than actually setting off real XHR requests inside tests.

dionrhys:
You may want to track the progress of https://github.com/Microsoft/TypeScript/issues/21414 as well.

That's an issue for fixing up these event type definitions in the Typescript project (and also mention the non-normative aspect of the spec).

fabriciomurta:
The OnProgress has that because it explicitly has it in MDN documentation: 
- [XMLHttpRequestEventTarget](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget)
- [OnProgress](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget/onprogress)

But the others are barely documented, and don't even say about an event argument.
- [OnLoad](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget/onabort)

dionrhys:
Hm, yeah. From your first link there I can see that `timeout`, `progress`, `loadstart`, `loadend` are specified as `ProgressEvent`. While `error`, `load`, `abort` are specified differently.

I think this is due to not updating the documentation on MDN myself. I'll have to see if it's possible to raise issues with MDN about these to clarify how these should be. In some of these cases, the event names are overloaded for `XMLHttpRequest` and behave differently to that same event name in other places (which is why `error` is apparently *"`UIEvent` if generated by a user interface element, `MediaRecorderErrorEvent` if generated by the MediaStream Recording API, and `Event` otherwise."* - I think these are all missing an extra overload for the case of when they're emitted on `XMLHttpRequestEventTarget` instances).

fabriciomurta:
Have you tried `Retyped.dom` package for Bridge? It tends to be more up-to-date and contain also non-solidly defined concepts (it also adopt some "non-living-standard" and non-normative good practices. Bridge.HTML5 is more tied to living standard only from MDN documentation.

dionrhys:
We are quite far off being able to migrate away from Bridge.Html5 at the moment because we have quite large codebases with dependencies spread over a lot of Bridge-based Nuget packages. All the code depends on Bridge.Html5 (things like [Bridge.React](https://github.com/ProductiveRage/Bridge.React), [ProductiveRage.Immutable](https://github.com/ProductiveRage/Bridge.Immutable), tens of in-house custom Bridge.React-based components that make use of JS APIs over Html5, an in-house Test Framework... the list goes on).

Untying ourselves from Bridge.Html5 will take quite a while because of the API surface area we have accepting/returning Bridge.Html5 classes. (I'm quite anxious about how we will migrate without hampering the pace of our project development in the interim)

I'm also unkeen on adopting Retyped because we'd then be relying on a proprietary service generating these Nuget packages without any insight into how they're built, no way to contribute code back into the project, and no means of recovering if support for Retyped is suddenly dropped. We would be reliant on outdated packages created by a third-party with no means of updating them ourselves.

_(These thoughts just reflect my own opinions - I don't speak with any kind of authority for our team here)_

In regards to this issue specifically, it appears `Retyped.dom` is built from TypeScript's own `lib.dom` (if I'm not mistaken?). I've linked https://github.com/Microsoft/TypeScript/issues/21414 above - these event types are also an issue on that project too.

fabriciomurta:
I expect mostly you'd have to change case of calls, because retyped retains lower/uppercase between C# and JavaScript (at least by default).

Thank you very much for your feedback. If you get trouble getting up and running with the unit tests, [ping us on gitter](https://gitter.im/bridgedotnet/Bridge).

geoffreymcgill:
The **Bridge.Html5** project is being archived as of Bridge 17.0 and has been moved to the [Frameworks](https://github.com/bridgedotnet/Frameworks/tree/v17) repo.

The priority for supporting the most up to date html dom api has shifted to [Retyped](https://retyped.com).

geoffreymcgill:
> was there a problem with this pull request that required additional work from the Bridge Team other than a review, merge and inclusion in a subsequent release?

Yes. Time to review the change, investigate why this change is required, compare to lib.dom.ts, write at least some basic tests, determine if this affects any other Bridge libraries that depend on Html5 (mostly jQuery, QUnit, & WebGL), determine if this affects Deck (it shouldn't, but still need to test), one final check, then merge the PR, and probably a few steps that I'm not thinking about. 

This process should likely take one of the Bridge team members about 2-4 hours to complete if there are no problems.

We've had to make some tough decisions on what projects to focus our time on going forward, and what projects we just do not have the resources (time + money) to support. In order to include this PR, I'd have to pull someone off other critical tasks, and unfortunately, that's not possible right now.

> I would ask you to reconsider reviewing and merging PRs from others who are willing to fix some of its problems.

We can't spend the time on **Bridge.Html5** PRs, but I have a proposal for @ProductiveRage and/or @dionrhys: we can add you as contributors to the [Frameworks](https://github.com/bridgedotnet/Frameworks) repo with **write** access. That will allow you to maintain **Bridge.Html5** as required and not rely on the Bridge team for updates.

If that's something you would be interested in, just let me know. 

I believe the PR should also be recreated against the Frameworks repo `v17` branch. It might be possible to cherrypick somehow from the original PR, but probably easier to just recreate.

ProductiveRage:
@geoffreymcgill That sounds like a very reasonable compromise. What would happen in terms of NuGet package publishing, would any fixes that we make become available to the public via the Bridge.Html5 NuGet package automatically through your build process or would we need to do something manually or would we effectively build private versions of the packages for our own purposes from the public GitHub repo?

geoffreymcgill:
@ProductiveRage sorry, but maybe I wasn't clear in my last post. Official **Bridge.Html5** Nuget packages could still be published by Bridge whenever you require, but @fabriciomurta would have to manually trigger our final publish to NuGet process.

If NuGet had project level auth tokens I'd just give you the rights to publish directly to NuGet, but I don't think that's possible.

ProductiveRage:
@geoffreymcgill - that made perfect sense, thanks!

It's a pity that NuGet is so rubbish with permissions like that. I'm still using a soon-to-be-deprecated NuGet.org login instead of a Microsoft account when I publish my packages and I wondered if one of the benefits of using Microsoft accounts would be more granular permissions.. but that was just optimism and not based on any facts, I haven't had a chance to confirm or deny yet (but I suspect that I'll be disappointed).

Is the [Discussion](https://github.com/orgs/bridgedotnet/teams/contributors/discussions) area something that you need to grant permissions to access? I get a 404 when I try to follow that link.

(@dionrhys - fyi)

geoffreymcgill:
> Is the Discussion area something that you need to grant permissions to access? I get a 404 when I try to follow that link.

You need to be part of the @bridgedotnet/contributors team. I will add you both, then you'll have **write** access to [Frameworks](https://github.com/bridgedotnet/Frameworks).

dionrhys:
Thank you for this @geoffreymcgill! It's good to be able to contribute back here while the lib.dom.ts changes are being mulled over on https://github.com/Microsoft/TypeScript/issues/21414. :smile:

For this issue in particular we'll write tests to ensure the browser implementations match the expected behaviour. And aside from that, see what we can do about MDN's docs.

PR #3384, opened 2018-01-26
Fix ProgressEvent Total and Loaded properties to be UInt64

dionrhys:
The `Total` and `Loaded` properties on `ProgressEvent` should be `ulong` (aka `UInt64`) according to the spec:
* https://xhr.spec.whatwg.org/#interface-progressevent
* https://heycam.github.io/webidl/#idl-unsigned-long-long

This change means we can support handling progress of AJAX requests exceeding 2 GiB.

https://deck.net/fc2d77a7b0978dde6d531feb7472a832
(**Note:** this also has to cast the event to `ProgressEvent` because of the issue in PR #3383)

**Expected result:**
`> System.UInt64`

**Actual result:**
`> System.Int32`

- Change `ProgressEvent`'s `Total` and `Loaded` from `int` to `ulong`

I haven't tested tests against this pull request because the Bridge.Test project is not open.


fabriciomurta:
At first, as this is changing a class' contract (public member's type), it would rather go to 16.8.0. So after we're done with 16.7.1 release, we'll handle this.

At the same time, I don't think this is a feature really explored and Bridge.Html5 is but a complement of Bridge, so maybe we may change our minds and get it into the next release (if so we'll update the pull request's milestone).

Thank you very much for your contribution!

fabriciomurta:
This may not be enough for it to work, unfortunately.
On Chrome, I get this, unless I'm not guessing the types correctly:
image

On dotnetfiddle, I've pasted the max safe integer value for reference and printed out the int/uint maximum values:

image

So while using ulong will increase the range of possible values, it seems a very big value provided from C# side won't work. Anyway, as far as the ProgressEvent variable is only filled by the client side (and never the other way around), ulong will happily contain any value that the client (at least chrome) may provide.

dionrhys:
I think this is just an inherent limitation in JavaScript that not all values in the 64-bit integer range can be represented because the only number type is the IEEE-754 Double-precision Floating Point `number` type.

This has previously come up in WebKit's own unit tests for `ProgressEvent`: https://bugs.webkit.org/show_bug.cgi?format=multiple&id=67537

I don't think Bridge can bend the game here - we just have to accept the limitations of the underlying technology.

fabriciomurta:
Agreed, all that might take us some time to accept this pull request are proper unit tests. Are you up for the challenge? Take a look on [this comment from Leonid about writing unit tests](https://github.com/bridgedotnet/Bridge/pull/2771#issuecomment-321916248).

More here: https://github.com/bridgedotnet/Bridge/wiki/Bridge.Test-Overview

This may prove real useful for you to integrate tests on a Bridge application you may happen to write.

dionrhys:
Oh, Bridge.Test is now publically-available? Is that new? I swear we couldn't build Bridge tests before because that project wasn't available!

I'll have a look at writing tests for these over the weekend :)

fabriciomurta:
Yes, you can now play around with tests! If you get in trouble trying to crack the test, [ping us on gitter](https://gitter.im/bridgedotnet/Bridge) we (or somebody else in the community) may just be around to help.

fabriciomurta:
Forgot to mention, development should be taken on Bridge's `dev` branch. I just moved your changes to the `dev` branch, please ensure you check this out before sending your unit tests' changes

fabriciomurta:
We'll have to move this fix's milestone to 16.8.0 as we couldn't merge it in time for the version closing.

fabriciomurta:
I see you still didn't have time to try with the unit tests, are you having a hard time cracking it or just couldn't invest time to?

dionrhys:
I haven't had time to invest into the unit tests yet. Other priorities have taken hold and the workaround of casting the type works in the meantime.

I may have time to write unit tests another day but I typically use Bridge at work so it's less appealing to me at the weekend!

fabriciomurta:
Alright, well, enter unit tests!

fabriciomurta:
The implementation will make it to the next Bridge version (see milestone, currently 16.8.0).

dionrhys:
Ah, thank you! 😄

PR #3400, opened 2018-02-03
Fixes BYTES_PER_ELEMENT constants for typed arrays.

### LINKED ISSUE #1373: Literal Multiplied by BYTES_PER_ELEMENT wrong result
When you multiply a literal by Float32Array.BYTES_PER_ELEMENT, only literal value is returned.
Also tried casting to int.
### Expected

8
### Actual

2
### Steps To Reproduce

``` csharp
public class App
{
    public static void Main()
    {
        int numb0 = Float32Array.BYTES_PER_ELEMENT;
        //echo -> 4
        int numb1 = 2 * Float32Array.BYTES_PER_ELEMENT;
        //echo -> 2
    }
}
```

kvee33:
Fixes #1373 .

- Modify the actual `BYTES_PER_ELEMENT` constants


kvee33:
third time's the charm

fabriciomurta:
I've just mentioned this pr in an issue we just created. We can't run the code to check whether the fix really works (to leave it as unit tests so we don't break it again in the future).

But on another thought, the classes affected by this issue shouldn't need the values fixed at all. As they are marked as `[External]`, they will always use the value of the class it represents in the browser (we make 'external' classes in the aim to represent something already present in the browser's space, like the typed arrays).

What I mean is, no matter what's on those variables, what determines its actual value in running code is what's in the browser!.. Check out this example:
https://deck.net/69e958e1186271c6130710b9239e3c63

And we get to another conclusion here: even with the fix you provided, as the actual value is ignored, issue #1373 won't be fixed (because it does not matter what value we add here). The problem is that the constant is not being output at all (check javascript line 12 in the deck above), when some math happens.

Anyway, your fix would be good for informational/documentation purposes, but it won't reflect in the code nor fix the issue, unfortunately. :(

fabriciomurta:
On a second thought, yes, your fix will effectively fix the issue. Seems like operations involving `const` variables would be "optimized out" by compiler, so Bridge only gets the computed value if we do `const * 2`, so it will never know it is referencing the constant.

Probably the best option here would be to transform that constant member into a readonly property, like this:
https://deck.net/cc0799d4a970f1ad9d2537bd64051c23

kvee33:
Yeah, I mentioned that [here](https://forums.bridge.net/forum/community/help/5187-fix-for-bytes_per_element-constants#post5187), probably should have double linked this. Oh well.
Anyway, I did a quick fix originally, your solution is definitely better, so I've made the relevant changes.

fabriciomurta:
Enter Unit Tests!

Please take your time to give the unit tests' commits a good look: We'd usually require them before accepting pull requests! 👍

fabriciomurta:
This pull request also fixes issue #3401.

kvee33:
Cool!
OK, I'll keep that in mind. Thanks, you've been really helpful! :]

PR #3412, opened 2018-02-11
[3324] GetType doesn't throw exception on null

### LINKED ISSUE #3324: GetType doesn't throw exception on null
GetType doesn't throw exception on null.

### Steps To Reproduce

https://deck.net/a2b9050ca543182d63cf105bd5f24dbd/third
https://dotnetfiddle.net/BYjLfl

```csharp
public class Program
{
    public static void Main()
    {
        float? v = null;
        Console.WriteLine(v.GetType().FullName);
    }
}
```

should throw exception.

vladsch:
Fixes #3324


fabriciomurta:
Branch and features rebased and pushed to `dev` (thus git does not detect the changes were been merged).

Issue #2887, opened 2017-06-29
Convert.ToString(new object()) shows Object not System.Object.


michaelcheers:
Convert.ToString(new object()) shows Object not System.Object.

### Steps To Reproduce

https://deck.net/fd3e53dc5a2bd366fd57ce6feb090939/third
https://dotnetfiddle.net/d3y1f1

```c#
public class Program
{
    public static void Main()
    {
        Console.WriteLine(Convert.ToString(new object()) == "System.Object" ? "Passed" : "Failed");
    }
}
```

fabriciomurta:
Pull request #3414 also fixes this.

fabriciomurta:
This issue has just been fixed by the above pull request that's been merged upstream. The fix will make it to the next Bridge release (see attached milestone).

fabriciomurta:
"Passing" on dev.deck.net: https://dev.deck.net/fd3e53dc5a2bd366fd57ce6feb090939/third

PR #3414, opened 2018-02-11
[3376] Complex Dictionary-Key gets serialized to [object Object]

### LINKED ISSUE #3376: Complex Dictionary-Key gets serialized to [object Object]
Hello,

I am using a dictionary with an object as key for accessing data with multiple different indexes. If I serialize this dictionary to JSONthe key gets serialized to [object, object] and can,t get deserialized.

Deck.NET Example
https://deck.net/56b2be737ccaf5bf0a8153fa5b170324

Code

```csharp
public class Program
{
    private class ComplexKey{
        public string Foo { set; get; }
        public string Bar { set; get; }
                
        public ComplexKey(string foo, string bar){
            Foo = foo;
            Bar = bar;
        }
        
         public override bool Equals(object obj)
        {
            return (obj is ComplexKey) && (((ComplexKey)obj).Foo.Equals(this.Foo)) &&
                    (((ComplexKey)obj).Bar.Equals(this.Bar));
        }

        protected bool Equals(ComplexKey other)
        {
            return string.Equals(Foo, other.Foo) && string.Equals(Bar, other.Bar);
        }

        public override int GetHashCode()
        {
            unchecked
            {
                return ((Foo != null ? Foo.GetHashCode() : 0) * 397) ^ (Bar != null ? Bar.GetHashCode() : 0);
            }
        }
    }
    
    
    public static void Main()
    {
        //create dictionary
        var testDic = new Dictionary<ComplexKey, string>();
        
        //add some data
        testDic.Add(new ComplexKey("Hello", "World"), "some data");
        testDic.Add(new ComplexKey("World", "Hello"), "some other data");
        
        //test access
        Console.WriteLine("access test with key: " + testDic[new ComplexKey("Hello", "World")]);
        
        //serialize
        var json = Newtonsoft.Json.JsonConvert.SerializeObject(testDic);
        Console.WriteLine("TestJson: \n\r" + json);
        //deserialize        
        var testDesDic = Newtonsoft.Json.JsonConvert.DeserializeObject<Dictionary<ComplexKey, string>>(json);
        //print back data
        Console.WriteLine("access test with key: " + testDesDic[new ComplexKey("Hello", "World")]);        
    }
}
```

vladsch:
Fixes #3376 

- ToString method mimics .NET (if ToString is not overrided then it displays type name)
- Nested type name uses + (plus) as separator between parent and own name

Another part of fix will be in Newtonsoft


fabriciomurta:
This also fixes #2887.

fabriciomurta:
Tests fixed and rebased with other commits. Github didn't detect the merge, but the diff is perfect.

PR #3417, opened 2018-02-15
[3408] Dictionary<T, _>.Values.CopyTo NotImplemented

### LINKED ISSUE #3408: Dictionary<T, _>.Values.CopyTo NotImplemented
Hi,
when I create a System.Collections.Dictionary<TKey, TValue> with TKey other than string or primitive type (I.E. any custom struct, nullable, DateTime or any custom class), you cannot use Values.CopyTo - it is not implemented inside System.Array.copyTo JS method. As a consequence is you cannot pass dictionary values into a List<T> collection in its constructor (pretty common use case). Dictionary with primitive keys works fine.

### Steps To Reproduce

https://deck.net/72b08c9e860dd1710473527bdce646f3

```csharp
public static void Main()
{
    var dic = new Dictionary<C, int>() ;
    dic.Add(new C(), 0);
    dic.Values.CopyTo(new int[1], 0);
    new List<int>(dic.Values);
}
class C { }
```

### Expected Result

Elements are copied without error.

### Actual Result

System.Exception: Uncaught System.NotImplementedException: copyTo

fabriciomurta:
Fixes #3408 .

- Changes System.Collections.Generic.Dictionary.getValue() client-side implementation to return an array of values when the dictionary keys are complex.
- Adds unit tests

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -12589,17 +12589,23 @@ Bridge.Class.addExtend(System.Boolean, [System.IComparable$1(System.Boolean), Sy
             },
 
             getValues: function () {
-                if (this.isSimpleKey) {
-                    var values = [];
+                var values = [];
+                var entry;
 
+                if (this.isSimpleKey) {
                     for (var i = 0; i < this.keys.length; i++) {
                         values.push(this.entries[this.keys[i]].value);
                     }
-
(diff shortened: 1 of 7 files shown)


fabriciomurta:
Changes to `Dictionary.GetKeys()` are in place, please take a look and let me know if there's still some pitfalls in!..

PR #3454, opened 2018-03-09
C#7 Is and Case patterns support

vladsch:


fabriciomurta:
This will be handled by the rebased version over #3461.

PR #3456, opened 2018-03-10
Read only structs support

vladsch:


fabriciomurta:
For the record, this is a C#7.2 (and newer) feature.
image

fabriciomurta:
Rebased and will be handled by #3462.

PR #3457, opened 2018-03-10
Expression body members support

vladsch:


fabriciomurta:
Pull req #3463 rebases this on Bridge v17 and merges. This is no longer necessary.

Issue #2455, opened 2017-03-13
Support C# 7


geoffreymcgill:
## TODO

- [x] is expression
- [x] switch pattern matching
- [x] local functions
- [x] expression-bodied members
- [x] binary literals
- [x] digit separators
- [x] ValueTuple
- [x] out variables
- [x] discards
- [x] throw Expressions

### Steps To Reproduce

https://dev.deck.net/ac2cca2ae5c12fe6eb21d15b175f31d6

```c#
using System;
using System.Collections.Generic;
using System.Linq;
using static System.Console;

//C# 7 is coming soon in VS 2017
//Try out some of the new features without having to install VS 2017 RC

//Source 1: http://www.c-sharpcorner.com/article/top-10-new-features-of-c-sharp-7-with-visual-studio-2017/
//Source 2: https://docs.microsoft.com/en-us/dotnet/articles/csharp/csharp-7

public class Program
{
	public static void Main()
	{
		//Small helper functions available only in the scope of parent function
		NestedFunctions();
		
		//Declare binary literals
		BinaryLiterals();
		
		//Declare out variable inside the method call
		OutVariables();
		
		//Declare tuples inline
		EnhancedTuples();
		
		//Access tuple members directly without using tuple name
		TupleDeconstruction();
		
		//Use expression bodied methods for properties, constructor and distructor
		ExpressionBodiedMembers();
		
		//Check for variable type in switch cases
		SwitchExpressions();
				
		//You can now throw exceptions inline in conditional expressions, null coalescing expressions, and some lambda expressions
		ThrowExceptionsInExpressions();
	}

	public static void NestedFunctions()
	{
		WriteFeatureHeader("Nested functions", "Small helper functions available only in the scope of parent function");
		
		void Add(int x, int y)
		{
			WriteLine($"Sum of {x} and {y} is : {x + y}");
		}

		void Multiply(int x, int y)
		{
			WriteLine($"Multiply of {x} and {y} is : {x * y}");
			Add(30, 10);
		}

		Add(20, 50);
		Multiply(20, 50);
	}

	public static void BinaryLiterals()
	{
		WriteFeatureHeader("Binary Literals", "Declare binary literals");
		
		//Represent 50 in decimal, hexadecimal & binary  
		int a = 50; // decimal representation of 50  
		int b = 0X32; // hexadecimal representation of 50 
		int c = 0B110010; //binary representation of 50  
		
		//Represent 100 in decimal, hexadecimal & binary  
[...]

PR #3461, opened 2018-03-11
c$7 "is" and switch case patterns

fabriciomurta:
Rebased over #3454.


fabriciomurta:
This is a C#7 feature disclosed in #2455.

PR #3462, opened 2018-03-11
c# 7.2 read only struct support

fabriciomurta:
This rebases #3456 onto Bridge v17.


fabriciomurta:
This is a feature disclosed in #3491.

PR #3463, opened 2018-03-11
V17 c sharp7 expression body

fabriciomurta:
Rebases #3457 on Bridge v17


fabriciomurta:
This is a feature disclosed in #2455.

PR #3464, opened 2018-03-11
[3440] override method not work correctly

### LINKED ISSUE #3440: override method not work correctly
override method not work correctly

### Steps To Reproduce

https://deck.net/69335d245c525361a707f982e64f49ff

```csharp
public class Program
{
    public static void Main()
    {
        System.IO.TextWriter Writer = new System.IO.StringWriter();
        Writer.ToString();
        
 	System.IO.StringWriter Writer1 = new System.IO.StringWriter();
        Writer1.ToString();
    }
}
```

### Expected Result

```js
    "use strict";

        main: function Main () {
            var Writer = new System.IO.StringWriter.ctor();
            Writer.ToString();

            var Writer1 = new System.IO.StringWriter.ctor();
            Writer1.ToString();
        }
    });
});
```

### Actual Result

```js
    "use strict";

        main: function Main () {
            var Writer = new System.IO.StringWriter.ctor();
            Bridge.toString(Writer);

            var Writer1 = new System.IO.StringWriter.ctor();
            Writer1.ToString();
        }
    });
});
```

vladsch:
Fixes #3440 

- if member is overriden then rules of base member (virtual) should have priority over class rules
- 
-


Tests/Runner/Batch3/code.js
```diff
@@ -29671,6 +29671,44 @@ Bridge.$N1391Result =                     r;
         }
     });
 
+    Bridge.define("Bridge.ClientTest.Batch3.BridgeIssues.Bridge3440", {
+        statics: {
+            methods: {
+                TestRulesForOverride: function () {
+                    var writer = new System.IO.StringWriter.ctor();
+                    Bridge.Test.NUnit.Assert.AreEqual("", Bridge.toString(writer));
```
fabriciomurta: This looks like what the issue complains about (should be something like `writer.ToString()`, not `Bridge.toString(writer)`.

If this is the expected behavior, then I don't understand what the changes in this pull request actually fix.

vladsch: It fixes name of overriden functions (override should have the same name as original member). Just see changes in io.js or encoding.js

fabriciomurta:
This issue, as far as the test cases are concerned, do no apply to Bridge 17. But the fix has several code involving "virtual" with "interface" which suggests some interaction between interfaces and classes having virtual-override-new handling has been implemented here -- which would not exist in Bridge 17 still.

PR #3465, opened 2018-03-11
CSharp7 Binary Literals and Digit Separators

vladsch:


fabriciomurta:
This is a bundle with two features mentioned in #2455.

PR #3473, opened 2018-03-17
V17 nuget upd

fabriciomurta:
Partially upgrades the Bridge solution's NuGet packages, avoiding version updates that would break build.

*Notice:* If building `Bridge.Test`, It may be necessary to check out `Test` repo's `v17_nugetUpd` branch in order to build the tests afresh.


fabriciomurta:
This is throwing:
```
Could not load file or assembly 'System.ValueTuple, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```
For Geoffrey

fabriciomurta:
As this is not working, another approach to upgrade NuGet packages will be necessary.

Issue #3478, opened 2018-03-20
DateTime.UtcNow returning wrong time


yctri:
I think the deck below (showing that DateTime.UtcNow displays the wrong time) may be a symptom of another problem under the hood. 

Given a previously-saved UTC time without timezone information, I believe we can usually reconstruct its .NET representation using:

```csharp
    // set DateTimeKind to be UTC.
    DateTime myUTCTime = DateTime.SpecifyKind(myTimeWithoutTimeZoneInfo, DateTimeKind.Utc);
```

The above used to work fine until updating to 16.8.2. Now, comparisons between myUTCTime and DateTime.UTCNow are returning a timezone offset. I noticed the issue below when attempting to reproduce this problem in Deck.

### Steps To Reproduce

https://deck.net/9a65e5cb6b9b80fe4f69d1bf9b3c69dd

```csharp
public class Program
{
    public static void Main()
    {
        Console.WriteLine("UTC time: " + DateTime.UtcNow);
        Console.WriteLine("Local time: " + DateTime.Now);
        Console.WriteLine("Local - UTC: " + (DateTime.Now - DateTime.UtcNow));
    }
}
```

### Expected Result

```js
UTC time: 03/21/2018 09:23:14
Local time: 03/20/2018 14:23:14
Local - UTC: -07:00:00
```

### Actual Result

```js
UTC time: 03/21/2018 04:23:14
Local time: 03/20/2018 14:23:14
Local - UTC: -07:00:00
```

geoffreymcgill:
Hi @yctri – Thanks for the report. We are investigating. 

We're looking into the Deck Share feature too. Looks like GitHub changed something with their Gist API that is now causing a failure in Deck.

geoffreymcgill:
@yctri – This issue has been fixed and pull-request has been submitted. Once testing and final review are complete, the pr will be merged into `v17` and will be included in the Bridge 17.0 release.

The problem was related to the timezoneoffset being added twice to the DateTime object.

If this is a supercritical defect for you, we might be able to make 16.8.3 release, or at least merge this pr into the `dev` branch (currently 16.8.3) which would allow you to build local packages.

Deck share links are still failing because GitHub changed something in their API, but appears to be undocumented. We have submitted a technical support ticket with GitHub.

fabriciomurta:
This expected result does not make sense to me:
```
UTC time: 03/21/2018 09:23:14
Local time: 03/20/2018 14:23:14
[...]

PR #3480, opened 2018-03-21
[3478] DateTime.UtcNow returning wrong time

geoffreymcgill:
Fixed multiple issues with DateTime, including #3478.


fabriciomurta:
The issues affecting DayOfYear property seems to be only triggered by South America (potentially south hemisphere countries), due to the different time zones set up.

PR #3487, opened 2018-03-23
Remove object.Keys() method

### LINKED ISSUE #3203: [Feature Request] May object.Keys() method be renamed to something less generic
Keys is very generic and very often used as an Enum in game engines/frameworks or UI toolkits, might this method be renamed to something less generic so it does not conflict with the possible enums?

https://github.com/bridgedotnet/Bridge/blob/f9f8ccda5433847f68ca73f67e34f4eb0a6f6978/Bridge/System/Object.cs#L47

geoffreymcgill:
Fixes #3203.

The Object.Keys() method is a not part of the C# **Object** class.


fabriciomurta:
As base for this merge was rebased (because its base was squashed), this will also need to be rebased, so the pull request will need to be created again.

PR #3490, opened 2018-03-23
[3203] May object.Keys() method be renamed to something less generic

### LINKED ISSUE #3203: [Feature Request] May object.Keys() method be renamed to something less generic

fabriciomurta:
Fixes #3203.

Branch needed to be rebased and this replaces pr #3487.

### DIFF
--- a/Bridge/System/Collections/Generic/Dictionary.cs
+++ b/Bridge/System/Collections/Generic/Dictionary.cs
@@ -37,7 +37,7 @@ namespace System.Collections.Generic
             get;
         }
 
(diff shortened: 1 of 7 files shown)


fabriciomurta:
Unit tests extending classes when they use `new (...) Key` will need review.

PR #3508, opened 2018-04-01
Support arrow local functions

vladsch:
Fixes # .

- 
- 
-


fabriciomurta:
The feature this implements is to support expression bodies in local functions (two C#7 features together).

PR #3520, opened 2018-04-05
[3511] IE11 error while accessing via interface

### LINKED ISSUE #3511: IE11 error while accessing via interface
See the following deck:

https://deck.net/e0004126083b25a802a938b627b5c5e0

IE11 prints `TypeError: Object doesn't support property or method 'Demo$IBar$1$Element$A'`

vladsch:
Fixes #3511


fabriciomurta:
Awesomesauce update! It works now!

PR #3526, opened 2018-04-10
ConstantPatternSyntax support

vladsch:
Support var keywords in case patterns


fabriciomurta:
This is a revision of features included in issue #2455.

PR #3576, opened 2018-05-16
[3546] New line emitted when calling multiline script

### LINKED ISSUE #3546: New line emmited when calling multiline script
New line emmited when calling multiline script.

Im not sure if this usecase is expected to be used in that way :)

### Steps To Reproduce

https://deck.net/5fbd63b8bd314550e325edd544f78599

```csharp
public class Program
{
    public static void Main()
    {
        string a = "foo";
        
        var b = Script.Call<string>(@"    
        (
        function(p){ return p + 'bar';}
        )", a);
        
        Console.WriteLine(b); // expected foobar
    }
}
```

### Expected Result

```js
    "use strict";

        main: function Main () {
            var a = "foo";

            var b =     
            (
            function(p){ return p + 'bar';}
            )(a);

            System.Console.WriteLine(b);
        }
    });
});


```

### Actual Result

```js
    "use strict";

        main: function Main () {
            var a = "foo";

            var b =     \r\n            (\r\n            function(p){ return p + 'bar';}\r\n            )(a);

            System.Console.WriteLine(b);
        }
    });
});


```

vladsch:
Fixes #3546


Compiler/Translator/Emitter/Blocks/InlineArgumentsBlock.cs
```diff
@@ -1158,7 +1160,7 @@ protected virtual void EmitInlineExpressionList(ArgumentsInfo argsInfo, string i
                     tempVars.Add(t, replacement);
                     tempMap[tempKey] = t;
                     return t;
-                }
+                }
```
fabriciomurta: Wonder where  those spaces come from.

Issue #3577, opened 2018-05-16
Module name escaping


AndreyChechel:
If a module name is not a valid identifier, it can't be used as is for naming a variable representing the module. In this case the identifier should be escaped.

### Steps To Reproduce

https://deck.net/894f8f874d20f647e65113f2a38df19d

```csharp
public class Program
{
    public static void Main()
    {
        var test = new MyModule.Test();
    }
}

[Name("MyModule")]
[Module("path/to/my/module")]
public class MyModule
{
    public class Test
    {
    }
}
```

### Expected Result

```js
Bridge.assembly("Demo", function ($asm, globals) {
    "use strict";

    require(["path/to/my/module"], function (path_to_my_module) {
        Bridge.define("Demo.Program", {
            $metadata : function () { return {"att":1048577,"a":2,"m":[{"a":2,"isSynthetic":true,"n":".ctor","t":1,"sn":"ctor"},{"a":2,"n":"Main","is":true,"t":8,"sn":"Main","rt":System.Void}]}; },
            main: function Main () {
                var test = new path_to_my_module.MyModule.Test();
            }
        });
        Bridge.init();
    });

    define("path/to/my/module", function () {
        var path_to_my_module = { };
        Bridge.define("MyModule", {
            $metadata : function () { return {"nested":[path_to_my_module.MyModule.Test],"att":1048577,"a":2,"m":[{"a":2,"isSynthetic":true,"n":".ctor","t":1,"sn":"ctor"}]}; },
            $scope: path/to/my/module
        });

        Bridge.define("MyModule.Test", {
            $kind: "nested class",
            $metadata : function () { return {"td":path_to_my_module.MyModule,"att":1048578,"a":2,"m":[{"a":2,"isSynthetic":true,"n":".ctor","t":1,"sn":"ctor"}]}; },
            $scope: path_to_my_module
        });
        return path_to_my_module;
    });

});
```

### Actual Result

```js
Bridge.assembly("Demo", function ($asm, globals) {
    "use strict";

    require(["path/to/my/module"], function (path/to/my/module) {
        Bridge.define("Demo.Program", {
            $metadata : function () { return {"att":1048577,"a":2,"m":[{"a":2,"isSynthetic":true,"n":".ctor","t":1,"sn":"ctor"},{"a":2,"n":"Main","is":true,"t":8,"sn":"Main","rt":System.Void}]}; },
            main: function Main () {
                var test = new path/to/my/module.MyModule.Test();
            }
        });
[...]

PR #3578, opened 2018-05-17
[3549] [3577]

### LINKED ISSUE #3549: "preventModuleName" argument value is not respected in non-external modules
If `[Module]` attribute is specified with a constructor argument `preventModuleName` set to `true`, the name of the module should not be emitted as a prefix in references to the module entities.

The issue is reproducible for non-external modules only. External modules process `preventModuleName` argument value correctly.

### Steps To Reproduce

https://deck.net/68bd3d55fd2684327594fd082313d8b0

```csharp
public class Program
{
    public static void Main()
    {
        var a = new test1.A();
        var b = new test2.B();
    }
}

[Bridge.Module(ModuleType.UMD, false, Name = "test1")]
public static class test1
{
    [Name("A")]
    public class A : Retyped.IObject
    {
    }
}

[Bridge.Module(ModuleType.UMD, true, Name = "test2")]
public static class test2
{
    [Name("B")]
    public class B : Retyped.IObject
    {
    }
}
```

### Expected Result

```js
// ... 
main: function Main () {
    var a = new test1.A();
    var b = new B();
}
// ...
```

### Actual Result

```js
// ... 
main: function Main () {
    var a = new test1.A();
    var b = new test2.B();  // Error:  "test2." is emitted regardless "preventModuleName" value
}
// ...
```

vladsch:
[3549] "preventModuleName" argument value is not respected in non-external modules
[3577] Module name escaping

Fixes #3549 and #3577


fabriciomurta:
Just synced up to `dev`. Shouldn't have broken anything, hopefully (builds and runs fine)

vladsch:
@AndreyChechel Please review additional commit with the fix

AndreyChechel:
Works correctly now, thanks!

fabriciomurta:
The exception output was not really helpful. Added the original exception's message to the output so errors during bridge.json file reading are properly handled.

The change by this issue implies that any new `Bridge.Contract.Module` instance should receive a reference to the translator's emitter. It seems the emitter is not available at the time or context that ReadConfig() lies within, or at least some parameter should be passed to the json deserialization command in order to effectively pass the emitter's reference to the constructor when type is `Bridge.Contract.Module`.

PR #3579, opened 2018-05-17
[3545] Reflection PropertyInfo.SetValue(obj, value) does not work on boxed value-type structs

### LINKED ISSUE #3545: Reflection PropertyInfo.SetValue(obj, value) does not work on boxed value-type structs
When in my application, for a generic PropertyGrid implementation, attempting to:
  a. Box a value-type structure into an Object
  b. Then change one of the properties through reflection
  c. Then Unbox it back into the value-type.

This should work, and does work in a C# Windows Console Application compiled with Visual Studio 2017.

In Bridge .NET it seems the change is not properly set inside PropertyInfo.SetValue in a silent failure. Perhaps lost in due to some internal boxing/unboxing mechanisms inside PropertyInfo.SetValue(), even though already boxed from the outside.

### Steps To Reproduce

https://deck.net/010f582b7cf0fa70c3fb3abf0722ea67

```csharp
public class Program
{
    // Author: Christian Lundheim
    // ISSUE: Reflection inconsistency with C# compiler on Windows
    
    // Change to 'class' and it will work as expected
    public struct Size
    {
        public int Width {get;set;}
        public int Height {get;set;}
    }

    public static void Main()
    {
        // Init value typed data-structure
        Size Test = new Size {Width = 10, Height = 20};
        
        // Box the value-type and change it through Reflection
        Object Boxed = Test;
        Boxed.GetType().GetProperty("Height").SetValue(Boxed, 1234);
        
        // Unbox it back
        Test = (Size)Boxed;
        
        // Does output:   Width=10, Height=20
        // Should output: Width=10, Height=1234
        // (correct output observed on C# Windows Console Application)
        Console.WriteLine($"Width={Test.Width}, Height={Test.Height}");
    }
}
```

### Expected Result

```
Width=10, Height=1234
```

### Actual Result

```
Width=10, Height=20
```

vladsch:
Fixes #3545


ChrML:
Great work :)

Bridge/System/Reflection/PropertyInfo.cs
```diff
@@ -1,3 +1,5 @@
+using Bridge;
```
fabriciomurta: Geoffrey decided not to add those to Bridge code anymore in favor of using full namespace.

PR #3341, opened 2018-01-02
Add support for new csproj format

Danielku15:
Fixes #3340.

Changes proposed in this pull request:

- Updates unit tests to year 2018
- Replace simple msbuild condition evaluation with full project loading. 
- Use Microsoft.Build nuget package to have new csproj support 
- Use VS2017 in AppVeyor

PR #3591, opened 2018-05-18
New cs proj format support

fabriciomurta:
Adds support for Bridge to parse and build projects using the new CSProj format.


fabriciomurta:
Couldn't push changes anymore to #3341. So pushed the merged version here so that it can be included into Bridge.

chenxustu1:
@fabriciomurta  is there a simple guide for how to use  new CSProj format with bridge?

Issue #3585, opened 2018-05-17
Incorrect message in exception thrown by IList for read-only collections


AndreyChechel:
The message thrown by `IList` for modification attempts of a read-only collection should be corrected.

### Steps To Reproduce

https://deck.net/752836565d7c92511823db09cb8903cc

https://dotnetfiddle.net/077OGB

```csharp
public class Program
{
    public static void Main()
    {
        var list = new List<int> { 1, 2, 3 };
        var roList = list.AsReadOnly();
        var ilist = (IList)roList;
		
        try
        {
            ilist.Insert(0, 0);
            Console.WriteLine("No Exception");
        }
        catch (Exception ex)
        {
            Console.WriteLine("[Error]: " + ex.Message);
        }
    }
}
```

### Expected Result

Console output:
```
[Error]: Collection is read-only.
```

### Actual Result

Console output:
```
[Error]: Collection was of a fixed size.
```

fabriciomurta:
Fixed and ensured the right message goes to the exception depending on the base type of list, discerning between fixed-length array and read-only lists.
The fix will make it to next Bridge release (see milestone).

Issue #3584, opened 2018-05-17
IList methods do not throw exceptions for fixed size arrays


AndreyChechel:
Instead of throwing **NotSupportedException**:
- `IList.Insert` does nothing
- `IList.Add` does nothing and returns `-1`
- `IList.Remove` removes an element
- `IList.RemoveAt` throws `TypeError: list.System$Collections$IList$removeAt is not a function at Function.IsIList`
- `IList.Clear` assigns `null` to existing elements

### Steps To Reproduce

https://deck.net/30cb3c32279b4fbe587048a8aa63b975

https://dotnetfiddle.net/AtaSgF

```csharp
public class Program
{
    public static void Main()
    {
        var arr = new int[] { 1, 2, 3 };
        var ilist = (IList)arr;
		
        try
        {
            ilist.Insert(0, 0);
            Console.WriteLine("No Exception");
        }
        catch (Exception ex)
        {
            Console.WriteLine("[Error]: " + ex.Message);
        }
    }
}
```

### Expected Result

Console output:
```
[Error]: Collection was of a fixed size.
```

### Actual Result

Console output:
```
No Exception
```

### See also
- [#3583] IList.IsFixedSize returns incorrect result

fabriciomurta:
After the fix, the fix also applied for #3585 implies in an incorrect exception message for the test case presented here.

fabriciomurta:
Fixed the fix, and the fix for the fix. So, all working. And the correction will make it to the next Bridge release (see milestone).

PR #3595, opened 2018-05-20
IList issues

### LINKED ISSUE #3583: IList.IsFixedSize returns incorrect result
`IList.IsFixedSize` returns `false` for fixed-size arrays.

### Steps To Reproduce

https://deck.net/ef9cd5d649b973a9860781fe07f33252

https://dotnetfiddle.net/4QV6L1

```csharp
public class Program
{
    public static void Main()
    {
        var arr = new int[] {1, 2, 3};
        Console.WriteLine("int[].IsFixedSize = " + arr.IsFixedSize);
        
        var ilist = (IList) arr;
        Console.WriteLine("IList.IsFixedSize = " + ilist.IsFixedSize);
    }
}
```

### Expected Result

Console output:
```
int[].IsFixedSize = True
IList.IsFixedSize = True
```

### Actual Result

Console output:
```
int[].IsFixedSize = True
IList.IsFixedSize = False
```

vladsch:
[3583] IList.IsFixedSize returns incorrect result
[3584] IList methods do not throw exceptions for fixed size arrays
[3585] Incorrect message in exception thrown by IList for read-only collections
Fixes #3583 #3584 #3585.


Bridge/Resources/Array.js
```diff
@@ -316,7 +316,7 @@
 
         checkReadOnly: function (obj, T, msg) {
             if (System.Array.getIsReadOnly(obj, T)) {
-                throw new System.NotSupportedException.$ctor1(msg || "Collection was of a fixed size.");
+                throw new System.NotSupportedException.$ctor1(msg || "Collection is read-only.");
```
fabriciomurta: Why was the string modified? Now it no longer matches the expected output that would happen from a native .NET code.
This is related to issue #3584.

fabriciomurta: It is possible to see the next issue, #3585, suggests this change, but it will be needed some checking to discern on the exception for test cases for issues #3584 (was of a fixed size) and #3585 (is read-only).

The summarized tests would be:
Expected/native: https://dotnetfiddle.net/p1JR3X
Bridge: http://deck.net/69d4ec3e1fb1592699bdaaf7d1882f34

Deck result on current state of the branch: 
```
Success: Type of exception is "NotSupported"
Failure: Expected exception message matches.
Success: Type of exception is "NotSupported"
Success: Expected exception message matches.
Statistics:
- Passing: 3 (75%)
- Failing: 1 (25%)
- Total: 4
```

PR #3602, opened 2018-05-22
StartsWith method does not compare strings correctly

### LINKED ISSUE #3593: StartsWith method does not compare strings correctly
Check sample:

### Steps To Reproduce

https://deck.net/3c0a66eae87385cde73a43f5cc5a433b
https://dotnetfiddle.net/Lg7R3U

```csharp
public class Program
{
    public static void Main()
    {
        Console.WriteLine("V".StartsWith("v", StringComparison.InvariantCultureIgnoreCase));
    }
}
```

### Expected Result

```
True
```

### Actual Result

```
False
```

geoffreymcgill:
Fixes #3593


fabriciomurta:
The escaping of the 'needle' string parameter during `(starts/ends)With()` was no longer required, notwithstanding, it incurred in issues whenever a needle string had characters taking "escape-aware" characters (such as dot). When escaped strings happened, the chunk split apart from the 'haystack' string never matched the actual string.
The needle string increased in length for each escaped character (backslash added), so then a simple comparison between the strings never matched as the escape characters were never took inconsideration, as the comparison no longer used regular expressions.

PR #3639, opened 2018-06-17
[3541] Metadata initialization fails when member metadata includes it…

### LINKED ISSUE #3541: Metadata initialization fails when member metadata includes its declaring type
Relates to Issue #3348:
> When a program uses modules, all entities will be wrapped with `require(["Module"], function(m) { .... } );` to resolve all dependencies first. The problem is that metadata must be initialized after entities are created. The require function changes normal flow of initialization and leads to errors.

The fix for #3348 helps when there are no members referring not initialized types. If such members exist, the issue is reproducible - it causes Bridge to add non-initialized namespace into a metadata class:
```js
// ... 
var $m = Bridge.setMetadata,
    $n = [System,Demo];      // Error: "Demo" is not defined
$m("Demo.App", function () { ... });
$m("Demo.Test", function () { ... });
```

### Steps To Reproduce

1. Create a solution with 2 Bridge projects named **Demo** and **PrinterModule**.
2. Add the following lines to **PrinterModule/App.cs** to define a module:
```cs
using Bridge;
using System;

namespace PrinterModule
{
    [Module(ModuleType.AMD, "PrinterModule")]
    public class Printer
    {
        public void Print(string msg)
        {
            Console.WriteLine(msg);
        }
    }
}
```
3. Reference **PrinterModule** project from **Demo** project.
4. Add the following lines to **Demo/App.cs**
```cs
using PrinterModule;

namespace Demo
{
    public class App
    {
        public static void Main()
        {
            var test = new Test();
            test.PrintInfo();
        }
    }

    public class Test
    {
        // Comment the field below to make the issue disappear.
        public static Test Instance = new Test();

        public void PrintInfo()
        {
            var p = new Printer();
            var type = GetType();

            p.Print(type.FullName + " contains " + type.GetMethods().Length + " method(s).");
        }
    }
}
```
5. Rebuild the solution
6. Add **require.js** script to the generated HTML file in **Demo** project:
```html
<!DOCTYPE html>

<html lang="en" xmlns="http://www.w3.org/1999/xhtml">
<head>
    <meta charset="utf-8" />
    <title>Demo</title>
    
    <script src="bridge.js"></script>
    <script src="bridge.console.js"></script>
    <script src="bridge.meta.js"></script>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.5/require.js"></script>
    <script src="PrinterModule.js"></script>
    <script src="Demo.js"></script>
    <script src="Demo.meta.js"></script>
</head>
<body>
</body>
</html>
```
7. Open that HTML file in browser.

### Expected Result

No errors. Console output:
```
Demo.Test contains 1 method(s).
```

### Actual Result

Console output:
```
Demo.Test contains 0 method(s).

2 Errors:
Uncaught ReferenceError: Demo is not defined
    at Demo.meta.js:6
    at Object.Bridge.assembly (bridge.js:3288)
    at Demo.meta.js:1

Uncaught TypeError: Cannot read property 'appendChild' of null
    at ctor.init (bridge.console.js:314)
    at new ctor (bridge.console.js:215)
    at Function.get [as instance] (bridge.console.js:22)
    at Function.logBase (bridge.console.js:90)
    at Function.error (bridge.console.js:122)
    at bridge.console.js:83
```

### See also
- https://github.com/Retyped/Retyped/issues/52#issuecomment-384375995

vladsch:
…s declaring type

Fixes #3541


fabriciomurta:
I spent some time trying to make an unit tests for this but found issues related to Bridge.qunit tests not including the tests which code inclusion was delayed due to the require() call to include the code.

PR #3660, opened 2018-07-09
[3540] Creating class instances of HTML elements inside template methods

### LINKED ISSUE #3540: Creating class instances of HTML elements inside template methods
When making a template method that can instantiate HTML elements.
Error is thrown in runtime at bridge.js:450 due to 'new type();' where type = HTMLDivElement.
HTMLDivElement as a native browser object is an Illegal constructor.

When invoking new HTMLDivElement() directly without template method, it translates to the following JS code: `document.createElement("div");`.
Thus issue only reproducable with template methods.

### Steps To Reproduce

https://deck.net/b7ce309fb4258993ab69b12d5b6585b4

```csharp
using Retyped;
using static Retyped.dom;

public class Program
{
    static void Main(string[] args)
    {
        // Built with Retyped
        // https://retyped.com
        
        // Create the Button the normal way
        var ButtonOK = new HTMLButtonElement();
        ButtonOK.innerText = "This button works";
        document.body.appendChild(ButtonOK);
        
        // Create the button using Template method
        // (ERROR generated here)
        var ButtonCrash = TemplateMethod<HTMLButtonElement>();	
        ButtonCrash.innerText = "This button would crash";
        document.body.appendChild(ButtonCrash);
    }
    
    static T TemplateMethod<T>() where T:class,new()
    {
        return new T();
    }
}
```

### Actual Result

```js
Same result as expected, issue is inside Bridge.createInstance(T);
```

vladsch:
Fixes #3540


ChrML:
Does this fix throw an InvalidOperationException or similar if the conditions for it to work is not present? Eg reflection is not enabled for that Retyped library.

PR #3715, opened 2018-09-16
[3714] Missing implicit conversion in base constructor call

### LINKED ISSUE #3714: Missing implicit conversion in base constructor call
The base constructor has a decimal parameter. If you pass 0, the value does not get converted to a decimal, leading to problems. Passing 0m does work.

### Steps To Reproduce

https://deck.net/8e74410d4c10713072c51b7ad4139356

```csharp
public class App
{
    public static void Main()
    {
        var test = new SubClass();
        Console.WriteLine(test.Value + test.Value);
    }
    
    public class SubClass : BaseClass
    {
        public SubClass() : base(0)
        {
            
        }
    }
    
    public class BaseClass
    {
        public BaseClass(decimal value)
        {
            this.Value = value;
        }
        
        public decimal Value { get; set; }
    }
}
```

### Expected Result
No exception

### Actual Result
System.Exception: TypeError: test.Value.add is not a function

### See Also
https://forums.bridge.net/forum/bridge-net-pro/bugs/5878-open-3714-missing-implicit-conversion-in-base-constructor-call

vladsch:
Fixes #3714

### DIFF
--- a/Compiler/Translator/Emitter/Blocks/ConversionBlock.cs
+++ b/Compiler/Translator/Emitter/Blocks/ConversionBlock.cs
@@ -156,7 +156,30 @@ namespace Bridge.Translator
                 var expectedType = block.Emitter.Resolver.Resolver.GetExpectedType(expression);
                 int level = 0;
 
-                if (expression.Parent is ParenthesizedExpression && expression.Parent.Parent is CastExpression)
+                if (expression.Parent is ConstructorInitializer ci && expectedType.Equals(rr.Type))
+                {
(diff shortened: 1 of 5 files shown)


fabriciomurta:
This was merged from a single integration branch to `dev`.
- bridgedotnet/Bridge@5d7fc47c05776e9ff1b531f9f5a7e226dee6eb86
- bridgedotnet/Bridge@ff4f7aa789bdd2dba47c739128d61ce9949b6957

PR #3716, opened 2018-09-16
[3709] Guid.ToString seems unneccessarily performance-inefficient

### LINKED ISSUE #3709: Guid.ToString seems unneccessarily performance-inefficient
Not a very important issue, but reporting it and you guys can choose what to do with it.

Been tracking the performance of my application lately to improve experience on lower-end pads/cellphones, and seems one of the major single points of slowdowns is caused by Guid.ToString. My application calls it not really more times than neccessary, just once each time a new Guid is generated in a loop with 30-40 items.

Looking at the calltree caused by Guid.ToString it seems there is significantly room for improvement / complexity reduction. A single call to Guid.ToString seems to:
- Visit 67 distinct javascript functions
- Construct a significant number of temporary items for immediate garbage disposal

See attached calltree.

guid tostring

vladsch:
Fixes #3709 

Test case https://deck.net/dd7a73ff50cc3a14cffff3fb623c75e0
Improvement time: from 2414 ms to 452 ms.


fabriciomurta:
This was merged from a single integration branch to `dev`.
- bridgedotnet/Bridge@6970a169f59ff99bb67abd2451eb2cbdfa440189

PR #3758, opened 2018-11-01
change clearTime in order to fix issue 3757

### LINKED ISSUE #3757: System.DateTime.Date returns incomplete date objects that cause Equals() to throw exception
Cannot use instance property 'Date' on System.DateTime. It seems that $clearTime doesn't populate kind and ticks causing it to throw exception.

### Steps To Reproduce

https://deck.net/b347ec5cb8266762cb3267c92ea892ea

```csharp
public class Program
{
    public static void Main()
    {
        var a = System.DateTime.Now.Date;
            var b = System.DateTime.Now;
            var c = System.DateTime.Now.Date;
			var bugCount = 0;
        
            Console.WriteLine($"a={a} b={b} c={c}");

            try {
                if (b.Equals(a)) 
                {
                    Console.WriteLine("[1] b should not be equal to a");
                    bugCount++;
                }
            } catch(Exception) 
            {
                Console.WriteLine("[2] equals throws exception");
                bugCount++;
            }
            
            try {
                if (a.Equals(b)) 
                {
                    Console.WriteLine("[3] a should not be equal to b");
                    bugCount++;
                }
            } catch(Exception) 
            {
                Console.WriteLine("[4] equals throws exception");
                bugCount++;
            }
            
            try {
                if (!a.Equals(c)) 
                {
                    Console.WriteLine("[5] a should be equal to c");
                    bugCount++;
                }
            } catch(Exception) 
            {
                Console.WriteLine("[6] equals throws exception");
                bugCount++;
            }

            Console.WriteLine($"ended {(bugCount == 0 ? "successfully" : "with bugs")}");       
    }
}
```

### Expected Result

```js
ended successfully
```

### Actual Result

```js
[2] equals throws exception
ended with bugs
```

d-p-y:
Fixes #3757 

- datetime misses kind and ticks causing Equals() to fail later

### DIFF
--- a/Bridge/Resources/Date.js
+++ b/Bridge/Resources/Date.js
@@ -1153,13 +1153,17 @@
                     dt.setUTCMinutes(0);
                     dt.setUTCSeconds(0);
                     dt.setUTCMilliseconds(0);
+                    dt.kind = 1;
                 } else {
                     dt.setHours(0);
                     dt.setMinutes(0);
                     dt.setSeconds(0);
                     dt.setMilliseconds(0);
+                    dt.kind = 2;
                 }
 
+                dt.ticks = this.getTicks(dt);
+                
                 return dt;
             },
 
--- a/Tests/Batch3/Batch3.csproj
+++ b/Tests/Batch3/Batch3.csproj
@@ -776,6 +776,7 @@
     <Compile Include="BridgeIssues\3700\N3713.cs" />
     <Compile Include="BridgeIssues\3700\N3714.cs" />
     <Compile Include="BridgeIssues\3700\N3728.cs" />
+    <Compile Include="BridgeIssues\3700\N3757.cs" />
     <Compile Include="BridgeIssues\3700\N3733.cs" />
     <Compile Include="BridgeIssues\TestBridgeIssues.cs" />
     <Compile Include="Utilities\BrowserHelper.cs" />
--- a/Tests/Batch3/BridgeIssues/3700/N3757.cs
+++ b/Tests/Batch3/BridgeIssues/3700/N3757.cs
@@ -1,0 +1,21 @@
+using Bridge.Test.NUnit;
+
+namespace Bridge.ClientTest.Batch3.BridgeIssues
+{
+    [TestFixture(TestNameFormat = "#3757 - {0}")]
+    public class Bridge3757
+    {
+        [Test]
+        public static void TestDate()
+        {
+            var a = new System.DateTime(2001, 2, 3, 4, 5, 6).Date;
+            var b = new System.DateTime(2001, 2, 3, 4, 5, 7);
+            var c = new System.DateTime(2001, 2, 3, 4, 5, 8).Date;
+            
+            Assert.True(!b.Equals(a)); //gist of issue: throws exception due to wrong DateTime.js $clearTime
+            //next two are safety guards that fix doesn't introduce new bugs
+            Assert.True(!a.Equals(b));
+            Assert.True(a.Equals(c));
+        }
+    }
+}


fabriciomurta:
Thank you for the pull request! Are you up for making the unit test for this fix?

That's easy! Just copy over another test case to this case (*N3757.cs*), replace the file texts that point to the issue, and add the code to test the issue, ensuring you use Assert.<method> to count something as working or not. Some useful methods are:
- `Assert.True`
- `Assert.AreEqual`
- `Assert.Throws`

You can search other test cases for how they are used.

Here's the last fixed issue's test case file you could base on: [N3733.cs on Bridge tag 17.4.0](https://github.com/bridgedotnet/Bridge/blob/v17.4.0/Tests/Batch3/BridgeIssues/3700/N3733.cs)

Looking forward to your test case!

d-p-y:
Hi Fabrício,
Thank you for checking it. I will provide unit test.

d-p-y:
Hi Fabrício,
I provided the test - it looks correct to me and it compiles. I followed [Bridge.Test wiki page](https://github.com/bridgedotnet/Bridge/wiki/Bridge.Test-Overview) but was not able to run Batch3. It failed in JS with "Bridge is not defined" so I added bridge.js to index.html but then it complained about Bridge.ClientTest.

fabriciomurta:
Hmm interesting, thanks for the feedback.

I have just tested a fresh clone of bridge repo and built `Bridge.Dev.sln`. It built successfully.
Then I scrolled down to the `Tests` folder, and set the `Runner` project as *Startup Project* (right click, and choose *Set as StartUp Project*).

Then voila. The tests page run when you build. You may build without debugging (in VS *Debug* > *Start Without Debugging*). I ensured to remove any local NuGet repositories so it pulled all from the web.

But you might of course want to route the generated Bridge Nuget packages to a local NuGet repo and cleanup/rebuild the project -- so that it grabs your built packages when linking the test projects.

To grab the built NuGet packages, just locate any .nupkg files within your repo after a successful build, then copy them over the local folder where you set your local NuGet repository; then next NuGet restore should pull them off.

You can use the 'find' tool if you have either cygwin or git bash shell. Here's what I get in the fresh repo after the first build:
```
find ./ -name '*.nupkg' -not -path './packages/*'
./Compiler/Build/bin/Release/Bridge.Clean.17.4.0.nupkg
./Compiler/Build/bin/Release/Bridge.Min.17.4.0.nupkg
./Compiler/Contract/bin/Release/Bridge.Contract.17.4.0.nupkg
./Compiler/Translator/bin/Release/Bridge.Compiler.17.4.0.nupkg
./PostBuild/bin/Release/Bridge.17.4.0.nupkg
./PostBuild/bin/Release/Bridge.Core.17.4.0.nupkg
```

Remember you need to remove the outdated copies of the packages within the `packages/` directory besides replacing the files in the local nuget repo.

This way you will run all tests integrated. You can then set a filter to show just your test. If the linked bridge version is still old, you'd notice your test failing (as it should without the fix!).

Hope this helps and does not sound too much way overcomplicated, just tried to cover most possible issues. Let us know if you give this a try and could have the tests running altogether. Would be nice to get feedback to improve the process.

d-p-y:
Fabrício,
thanks for explaining how to run those tests. Your comment would nicely fit into some kind of 'How to run bridge tests' section in readme.md :) Anyway - my test fails without Date.js fix. With fix applied it passes correctly (and all the other pass as well).

fabriciomurta:
Thanks for the feedback, wouldn't make much sense publishing this step-thru if it didn't actually work for you. As we frequently do this here, it may look like obvious to us and I didn't even think there was nothing about how to run it. But it should indeed be written somewhere.

I've just added this wiki entry for running the tests. I believe it should be pretty easy to follow and cover some pitfalls for people developing Bridge too. Suggestions, as always, are welcome!
[Bridge wiki page: Build & run Bridge self Tests](https://github.com/bridgedotnet/Bridge/wiki/Build-&-run-Bridge-self-Tests)

fabriciomurta:
Just a reminder, this pull request can potentially fix issue #3773, so check it up when reviewing the pull request.

d-p-y:
Fabrício,
Your wiki entry for building tests is detailed and helpful. I'm convinced that since now it will be clear for anybody on how to run tests. Thank you!

PR #3834, opened 2019-02-17
[3823] Checking if an object is decimal always results as true

### LINKED ISSUE #3823: Checking if an object is decimal always results as true
https://deck.net/3b219c3202f275e7bbd6cf19272ce72d

vladsch:
Fixes #3823


fabriciomurta:
This one looks good, the fix is ready to be released.

PR #3837, opened 2019-02-18
[2522] Implement Type.IsPrimitive

### LINKED ISSUE #2522: Implement Type.IsPrimitive

vladsch:
Fixes #2522

### DIFF
--- a/Bridge/System/Type.cs
+++ b/Bridge/System/Type.cs
@@ -417,5 +417,11 @@ namespace System
             [Bridge.Template("Bridge.Reflection.isValueType({this})")]
             get;
         }
+
+        public extern bool IsPrimitive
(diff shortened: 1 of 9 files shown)


Bridge/System/Type.cs
```diff
@@ -417,5 +417,11 @@ public abstract class Type
             [Bridge.Template("Bridge.Reflection.isValueType({this})")]
             get;
         }
+
+        public extern bool IsPrimitive
+        {
+            [Bridge.Template("Bridge.Reflection.isPrimitive({this})")]
```
fabriciomurta: This is not desired but would require a significant amount of refactoring to allow (if currently possible) being without template.

PR #3842, opened 2019-02-19
[3720] Enum.ToObject is missing

### LINKED ISSUE #3720: Enum.ToObject is missing
The ToObject(Type, Object) method converts the integral value value to an enumeration member whose underlying value is value.

This conversion method returns a value of type Object: [MSDN](https://docs.microsoft.com/en-us/dotnet/api/system.enum.toobject?view=netframework-4.7.2#System_Enum_ToObject_System_Type_System_Object_)

Used [this](https://deck.net/fca682377389f97195846bde3944a980) as workaround.

vladsch:
Fixes #3720


Bridge/System/Enum.cs
```diff
@@ -45,6 +45,9 @@ public abstract class Enum : ValueType, IComparable, IFormattable
 
         [Bridge.Template("System.Enum.format({this:type}, {this}, {format})")]
         public extern string ToString(string format, IFormatProvider formatProvider);
+
+        [Bridge.Template("System.Enum.toObject({enumType}, {value})")]
```
fabriciomurta: Undesired template. But it would take a considerable effort to convert the whole class in order to use a non-template approach.

Issue #3865, opened 2019-03-01
Array of Array haven't assembly


miladsadeghi72:
i using assembly for Serialize and Deserialize  objects
I got accidental this exception on sample tests

https://deck.net/db0da0f53bd7d7736ed74ec40856db31

```csharp
public class Program
{
    public static void Main()
    {
            try
            {
                Console.WriteLine(typeof(string[][]).Assembly.FullName);
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex.Message);
            }
        
    }
}
```

### Expected Result on console

mscorlib

### Actual Result on console

Bridge.Reflection.getTypeAssembly(...) is undefined

fabriciomurta:
Fixed! Thanks for not only reporting the issue, but also proposing a fix, and also understanding why we took the one from @vladsch! We're looking forward for further contributions from you in the future!

The fix will make it to the next Bridge release, version number according to the milestone attached to this issue.

geoffreymcgill:
Original issue by @miladsadeghi72:

i using assembly for Serialize and Deserialize  objects
I got accidental this exception on sample tests

https://deck.net/db0da0f53bd7d7736ed74ec40856db31

```csharp
public class Program
{
    public static void Main()
    {
            try
            {
                Console.WriteLine(typeof(string[][]).Assembly.FullName);
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex.Message);
            }
        
    }
}
```

### Expected Result on console

mscorlib

### Actual Result on console

Bridge.Reflection.getTypeAssembly(...) is undefined

PR #3866, opened 2019-03-01
_FIX #3865

miladsadeghi72:
#3865

### DIFF
--- a/Bridge/Resources/Reflection.js
+++ b/Bridge/Resources/Reflection.js
@@ -269,7 +269,7 @@
 
         getTypeAssembly: function (type) {
             if (type.$isArray) {
-                return type.$elementType.$assembly;
+                return Bridge.Reflection.getTypeAssembly(type.$elementType);
             }
 
             if (System.Array.contains([Date, Number, Boolean, String, Function, Array], type)) {


fabriciomurta:
Hello @miladsadeghi72, and thanks for the contribution!

You fix looks just right, there shouldn't anything else to be added regarding the fix. It's simple and effective. There are though, a couple issues with your pull request:
1. You didn't sign the CLA, we can't accept it even if it was perfectly fine.
2. You didn't provide unit tests for your fix, we simply can't merge fixes before they have an unit test -- well, when they can have an unit test, which is pretty much 100% of the case when a deck is possible to reproduce the issue.

In the other hand, it seems @vladsch took over the task and circumstantially the fix was but the same, and I don't think there was any other better way to fix this. So this could come in very handy for you, please take a look at vlad's pull request #3868 and see what I mean.

Once you're viewing his #3868 pull request, click the [Files Changed](https://github.com/bridgedotnet/Bridge/pull/3868/files) tab to see the changes. Essentially, the unit test is in the introduced file to the **Tests/Batch3** project: Tests/Batch3/BridgeIssues/3800/N3865.cs. You can always use another test code to create yours.

If you're a little lost with all this "unit tests" stuff, I suggest you follow this wiki post on running the tests:
https://github.com/bridgedotnet/Bridge/wiki/Build-&-run-Bridge-self-Tests

You should be able to run it, and see a whooping amount of 200,000 unit tests run on your browser. Using latest google chrome is advised, as we didn't have means to fix some tests on select browsers.

As a pull request fixing this in the format we require is already present, I'm closing this one. But please don't get the wrong idea, we really do appreciate you taking your time in contributions to the project and sharing your knowledge with us. We're looking forward to your next contribution!

miladsadeghi72:
Thanks bro but its not important, tnx for fix issue

miladsadeghi72:
I using this on real project so its the real test for bridge
I just want to bridge be safe, not important who fix or who accepted

PR #3916, opened 2019-04-16
[3915] Guid parsing is slow

### LINKED ISSUE #3915: Guid parsing is slow
### See Also
https://forums.bridge.net/forum/community/help/6103-newtonsoft-json-jsonconvert-deserializeobject-is-slow

vladsch:
Fixes #3915 

If use the following test case as base then deserialization time was improved (on my local machine) from 5570 ms to 3089 ms

https://deck.net/b672e0b116e7d773b6d0a942535726d3


vladsch:
I added two extern helpers classes for native RegExp. They are marked as internal, so it is not visible outside Bridge assembly

rootflood:
https://dev.deck.net/afdabbf583bafa1f55482d93111c92d1

fabriciomurta:
Thanks for the Deck, @rootflood. Your approach has similar deserialize performance to what the optimization reached, yet it does a noticeable faster serialization from ~2800ms to ~1600ms on my side.

The optimization introduced by this pull request is irrelevant to the deck you provided -- running it either in a patched or non-patched version of Deck results in similar performance -- so it may be used as a means to workaround the slow performance in Bridge 17.7.

PR #3961, opened 2019-05-16
[3962] IAsyncResult implementation for Task

vladsch:
https://forums.bridge.net/forum/community/help/6180-updates-and-problem-encountered

See #3962


Bridge/System/Threading/Tasks/Task.cs
```diff
@@ -39,6 +39,16 @@ public class Task : IDisposable, Bridge.IBridgeClass
             get;
         }
 
+        public object AsyncState
+        {
+            get;
+        }
+
+        public bool CompletedSynchronously
```
fabriciomurta: This is not [how it is defined in .net](https://github.com/microsoft/referencesource/blob/master/mscorlib/system/threading/Tasks/Task.cs#L1619), and allows code that won't work on native .NET to work on Bridge (implicit cast, accessibility).

fabriciomurta: Fixed with e5345a276b8da72ff1e52a5e5cba93fee538441f.

fabriciomurta:
Fixed the visibility and added unit tests, now everything looks fine to go to the next Bridge release.

PR #3966, opened 2019-05-22
Implement System.Array.LongLength and System.Array.Copy for long indexes

johnwason:
Implements Array.LongLength and System.Array.Copy for use with longs instead of ints.

### DIFF
--- a/Bridge/Resources/Array.js
+++ b/Bridge/Resources/Array.js
@@ -1255,6 +1255,9 @@
             }
 
             return arr || result;
+        },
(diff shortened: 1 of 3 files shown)


fabriciomurta:
The pull request looks really clean and tidy! I have created specific issues tailored for this pull request, as we do want the implemented feature to be in our changelogs when the release goes live -- and we base the changelog off issues, mainly, although we also tag pull requests after the version they go in.

In this case, I've created issues #3967 and #3968 to log the longlength property and array copy, respectively.

PR #3972, opened 2019-05-23
[3964] string.IndexOfAny is implemented incorrectly

### LINKED ISSUE #3964: string.IndexOfAny is implemented incorrectly
A description of the issue.
`string.IndexOfAny()` is implemented incorrectly.
The order of the input character array affects the outcome and it shouldn't.
For example, if you switch `H` and `d` in the character array in the example below, you'll notice that the example outputs `0` as expected instead of `11`.

### Steps To Reproduce

https://deck.net/37dbfd6bb322075afc63a3a31b8366d2

https://dotnetfiddle.net/w8cu4N

```csharp
public class Program
{
    public static void Main()
    {
        char[] test = {'d','H'};
        var msg = "Hello, World!";
        
        Console.WriteLine(msg.IndexOfAny(test));
    }
}
```

### Expected Result

The console should output `0` here.

### Actual Result

The console outputs `11`.

vladsch:
Fixes #3964

### DIFF
--- a/Tests/Batch3/BridgeIssues/3900/N3964.cs
+++ b/Tests/Batch3/BridgeIssues/3900/N3964.cs
@@ -1,0 +1,34 @@
+using Bridge.Test.NUnit;
+
+namespace Bridge.ClientTest.Batch3.BridgeIssues
+{
+    [TestFixture(TestNameFormat = "#3958 - {0}")]
+    public class Bridge3964
+    {
+        [Test]
+        public static void TestIndexOfAny()
+        {
+            char[] test = { 'd', 'H' };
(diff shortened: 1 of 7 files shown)


Tests/Batch3/BridgeIssues/3900/N3964.cs
```diff
@@ -0,0 +1,34 @@
+using Bridge.Test.NUnit;
+
+namespace Bridge.ClientTest.Batch3.BridgeIssues
+{
+    [TestFixture(TestNameFormat = "#3958 - {0}")]
```
fabriciomurta: Ops, it's 3964!

PR #3990, opened 2019-06-05
Fix ReadStream for resources larger than 1MB - fixes issue #3989

### LINKED ISSUE #3989: Resource larger than 1MB can not be extracted by dependent project
I have Project A and Project B where B references A.

Project A has a web font resource built into it (using the [resources](https://github.com/bridgedotnet/Bridge/wiki/global-configuration#resources) bridge.json option).

Project A builds correctly and the resource appears in the dll. Project B will not build, there is an exception thrown while Bridge tries to extract the resource out of the dll -

> **System.ArgumentException:** Offset and length were out of bounds for the array or count is greater than the number of elements from index to the end of the source collection.

I ran the Bridge resource-extracting code in isolation and reproduced the issue:

image

The approach of looping and reading chunks of data used to be common (eg. [https://stackoverflow.com/a/1080445](https://stackoverflow.com/a/1080445)) but the stream's Read method's summary comment mentions:

> Exceptions:
>   T:System.ArgumentException:
>   The sum of offset and count is larger than the buffer length.

.. and this seems to be the condition that we're falling into (file is approx 1.5MB, the second read loop will specify an offset of 1024 * 1024 and a count of 1024 * 1024 and the sum of these two values is larger than the size of the buffer).

.NET 4 made this sort of operation simpler when the CopyTo method was added to the Stream type. That could make this code simpler and more reliable.

ProductiveRage:
Fixes #3989 

I haven't added any tests for this - I wasn't sure where would be best to put one and whether it would be best to do a full integration test that reads a larger-than-1MB resource from a DLL included in the project or if it should just try to read from a stream that is larger than 1MB. I can add tests if you think that it would be helpful and could offer some guidance.

### DIFF
--- a/Compiler/Translator/Translator/Translator.Resources.cs
+++ b/Compiler/Translator/Translator/Translator.Resources.cs
@@ -1356,32 +1356,16 @@ namespace Bridge.Translator
 
         protected byte[] ReadStream(Stream stream)
         {
+            if (stream is MemoryStream memoryStream)
+                return memoryStream.ToArray();
+
             if (stream.Position != 0)
-            {
                 stream.Seek(0, SeekOrigin.Begin);
-                stream.Position = 0;
-            }
-
-            byte[] bytes = new byte[stream.Length];
-
-            int numBytesToRead = (int)stream.Length;
-            int numBytesRead = 0;
-
-            var chunkLength = 1 * 1024 * 1024;
-            if (chunkLength > numBytesToRead)
+            using (memoryStream = new MemoryStream())
             {
-                chunkLength = numBytesToRead;
+                stream.CopyTo(memoryStream);
+                return memoryStream.ToArray();
             }
-
-            do
-            {
-                int n = stream.Read(bytes, numBytesRead, chunkLength);
-
-                numBytesRead += n;
-                numBytesToRead -= n;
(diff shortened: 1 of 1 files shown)


fabriciomurta:
I believe integration tests to be the way to go with this. Maybe instead of having a 1Mb-big file sitting around, we could generate it at build-time, so we don't need to commit blobs. If you think a way into that, please don't hesitate to add!

p.s.: I know we commit lots of blobs in Bridge repo, but this was needed at some point to enforce unit tests, which should soon be reviewed so they are again built from the testing endpoint. So the goal is to stop committing those test output files at some point.

Tip: branch off to make the fixes you send. This will pretty much lock up your master branch until we review and merge the pull request, if you created a branch (say, **Issue3989** as we usually do here), you could just let it sit and focus on other fixes you might implement in the meanwhile.

ProductiveRage:
@fabriciomurta - your branching suggestion is very sensible, I will bear it in mind in the future!

I'll think about adding a test for the ReadStream method, the simplest approach would be to give it a memory stream that has more than 1MB of data - that way, there are no blobs checked into the repo and no possibility of file access problems with build. It's not a full end-to-end integration test for resource extraction but it would be a good start!

PR #4035, opened 2019-08-02
Avoid allocations in string.IsNullOrWhitespace calls

ProductiveRage:
We were doing some browser memory profiling / optimisation here and saw the following numbers that surprised us:

image

.. it didn't seem like "isNullOrWhiteSpace" should be something that should allocate if it's just comparing a string using a reg ex - the reason looks to be because the reg ex is instantiated for every call, rather than being created once and reused.

The reg ex in question is in System.Char and I noticed the pattern in some other methods in there, so I changed them all.

I've used the Runner project to execute the unit tests in the browser and 12 failures are reported but the same failure were reported before my changes.


ProductiveRage:
I thought that I should add some more info in case you want to confirm the impact of the changes. If I run the following JS:

```
setTimeout(
	function () {
		console.log("Starting..");
		var s = "abc";
		for (var i = 0; i < 1000000; i++) {
			System.String.isNullOrWhiteSpace(s);
		}
		console.log("DONE!");
	},
	3000
);
```

.. and use the "start recording heap profile" functionality in Chrome's Memory tab then before the changes the results look like this:

image

After the changes, the results look like this:

image

fabriciomurta:
Hello @ProductiveRage, have you used private/proprietary tooling to extract these results? I don't seem to find a view like your screenshots in either chrome or firefox developer tools.

I've logged issue #4036 and am cross-pointing issue and pull requests -- so that this gets to changelog when merged upstream.

Thanks for the finding and contribution!

ProductiveRage:
Sorry, I didn't leave the most detailed instructions there!

I used Chrome. In the Dev Tools, there is a Memory tab. In that, I selected "Allocation instrumentation on timeline" and checked the "Record allocation stacks (extra performance overhead)" option. Then I reloaded the page and pressed the record button at the top left of the Memory tab and waited for the page to do its work (I went to the Console tab and waited for the "DONE!" message). Then I went back into the Memory tab and clicked the record button again to stop the recording.

This produces a snapshot. The default view is "Summary", which I changed to "Allocation" to see the information attached to this case.

image

image

fabriciomurta:
Got it, thanks for the quick howto! :)

fabriciomurta:
Hi! I'm closing your pull request.
But falter not, I've cherry picked your commit, and squashed your second commit with also another change from myself in which we turned the `$<variable>` local variables you created into `_<variable>`.

The commits rebuild to keep a clean history, and the variable renaming to keep using the conventions I've seen around in other js files within *Bridge/resources* directory.

PR #4048, opened 2019-08-21
[953] Pure attribute to avoid clones

### LINKED ISSUE #953: Pure attribute to avoid clones
In the compiled code, I see a lot of calls to `$clone()`. I suggest you add an attribute like `PureAttribute` to specify that a method is not going to modify the parameter.

    [AttributeUsage(AttributeTargets.Parameter, Inherited = false, AllowMultiple = false)]
    sealed class PureAttribute : Attribute
    { }

Example:

    void Clear([Pure] Color fillColor) { ... }

    [Pure]
    void DrawRectangle(..., Color strokeColor, Color fillColor) { ... }

When the `PureAttribute` is encountered on a parameter, it would pass the parameter directly to `Clear` rather than making a clone. Alternatively, if encountered on a method, all parameters are passed directly.

stepan-beresnev:
Fixes #953

### DIFF
--- a/Compiler/Contract/Helpers.cs
+++ b/Compiler/Contract/Helpers.cs
@@ -451,6 +451,11 @@ namespace Bridge.Contract
                     return;
                 }
 
+                if (Helpers.IsPure(expression, block.Emitter))
+                {
+                    return;
+                }
+
                 if (writeClone)
                 {
                     Helpers.WriteClone(block, insertPosition, nullable);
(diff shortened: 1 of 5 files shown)


geoffreymcgill:
@stepan-beresnev Thanks for submitting this PR. We will review ASAP.

yard:
Hey guys! When can we expect this to get mainlined?

ChrML:
Good idea. However I think this should be reviewed in more details, because this is not standard .NET behavior, and merging it could cause problems down the line with people using this feature.

A few points to consider:

1. Usually a struct that you don't want cloned could be a class instead.

2. As of C# 7.2-8.0 there has been done quite a lot of work in the language standards and compiler to deal with this:

a) Immutable structs can be declared as "readonly struct" to enforce immutability and tells the compiler it can safely be passed around as a reference because it can't change (no cloning required).

b) Properties and methods in a struct that don't change it can also be declared as "readonly". This avoids the need to clone the struct before invoking that method/property, because compiler knows it won't change it.

c) If you want to mutate a struct given as an argument to a method, you can pass it with the "ref" keyword. Tells the compiler not to clone the struct, but pass it as a reference which the method may or may not modify, similar to the example in the PR.

d) If you don't want to mutate, nor have to clone it, on a per-method basis that accepts a struct parameter, use the "in" keyword.


My opinion is that it's not a good idea to merge this, but instead try to implement these new C# features that serve the same purpose.

PR #4069, opened 2019-10-05
Fix bug with "hasOwnProperty" dictionary key

Zarel:
Using `hasOwnProperty` as a dictionary key causes Bridge.NET to crash:

https://deck.net/ff2b4e49b1057c8ae99c566343b533de

The problem is that it's compiling to `obj.hasOwnProperty(key)`
when it should be compiling to `Object.prototype.hasOwnProperty.call(obj, key)`.

This PR fixes that crash.

### DIFF
--- a/Bridge/System/Object.cs
+++ b/Bridge/System/Object.cs
@@ -24,6 +24,7 @@ namespace System
         public virtual extern object ValueOf();
 
         [Bridge.Convention(Bridge.Notation.CamelCase)]
+        [Bridge.Template("Object.prototype.hasOwnProperty.call({this}, {v})")]
         public virtual extern bool HasOwnProperty(object v);
 
         [Bridge.Convention(Bridge.Notation.CamelCase)]


Bridge/System/Object.cs
```diff
@@ -99,4 +100,4 @@ public static class ObjectExtensions
         [Bridge.Template("Bridge.is({obj}, {T})")]
         public static extern bool Is<T>(this object obj);
     }
-}
\ No newline at end of file
+}
```
Zarel: image

GitHub's online editor seems to have added a trailing newline; I can revert that if people care.

fabriciomurta: I guess this is alright, often this likes to change and that's but a minor thing.

fabriciomurta:
The pull request has been accepted and merged. This just does not reflect in github because I rebased your branch over Bridge's. Your commit became: 46e972f24e3a63888bb3138c4b523085a8ecfb4c

If you're curious about the unit tests, its commit became 0e2227bbd7a617ff1008f13da2b4bbdd0a4eae1b.

PR #4097, opened 2019-12-16
[3322] Clone for Structs in Generics

### LINKED ISSUE #3322: Structs don't copy when assigned by out when assignee is ref.
Structs don't copy when assigned by out when assignee is ref.

### Steps To Reproduce

https://dev.deck.net/93c8466cda41dc6279076d52b4c50a5a/third
https://dotnetfiddle.net/IpCDZK

```csharp
public class Program
{
    public struct Int
    {
        public int Value;

        public void Increment() => Value++;
    }
    
    public static void Main ()
    {
        Int? num = new Int(), num2;
        Set(out num2, ref num);
        num.Value.Increment();
        Console.WriteLine(num2.Value.Value + " should be 0.");
    }
    public static void Set<T>(out T target, ref T input) => target = input;
}
```

stepan-beresnev:
Fixes #3322

### DIFF
--- a/Bridge/Resources/.generated/bridge.js
+++ b/Bridge/Resources/.generated/bridge.js
@@ -35,6 +35,10 @@
             }
         },
 
+        rValue: function (instance) {
+            return instance && instance.$clone ? instance.$clone() : instance;
+        },
+
         toString: function (instance) {
             if (instance == null) {
                 throw new System.ArgumentNullException();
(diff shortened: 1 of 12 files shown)


stepan-beresnev:
AppVeyor build fails with an error, which seems to be not related to the pull. Had the same issue locally, it fixed for me after updating Microsoft.Build, Microsoft.Build.Framework and Microsoft.Build.Utilities.Core to the latest version (16.4.0).

PR #4107, opened 2020-02-10
Runs BridgeNetCompilerTask only if sources changed

slavikn84:
Fixes the issue described in https://forums.bridge.net/forum/community/help/6093-bridge-compiler-running-on-every-project-in-solution-even-if-no-changes.

- Added inputs and outputs to the Task, so that msbuild will only trigger the task if the source c# files changed.

### DIFF
--- a/.build/files/Bridge.Min.targets
+++ b/.build/files/Bridge.Min.targets
@@ -19,7 +19,7 @@
     <PrepareForRunDependsOn>$(PrepareForRunDependsOn);$(BridgeNetCompilerTarget)</PrepareForRunDependsOn>
   </PropertyGroup>
 
-  <Target Name="BridgeNetCompilerTask">
+  <Target Name="BridgeNetCompilerTask" Inputs="$(OutputPath)$(AssemblyName).dll" Outputs="$(OutputPath)bridge\$(AssemblyName).js">
     <Message Text="Bridge started" Importance="high" />
 
     <BridgeCompilerTask


geoffreymcgill:
Are you using Visual Studio?

This PR is unlikely to get accepted as is because we cannot reproduce a scenario in which this change does anything different than the default behaviour.

A few other issues:

1. The outputs path assumes Bridge JS will always be output to `$(OutputPath)/bridge/`
2. The outputs path assumes Bridge JS will always be output as its Assembly name, matching `$(AssemblyName).js`, which is also a setting depending on **Bridge.json** or assembly/attributes settings.

Implementing this change makes several assumptions about the the output folder and output file name, and hard-codes those assumptions. This would ignore settings from **Bridge.json**.

slavikn84:
I compile from command line (msbuild), not from Visual Studio.
It's weird that you can't reproduce it - it happens to me 100% of the time,
and also happened to the guy who wrote the post. What I see is, every time
I compile (from the command line), the BridgeNetCompilerTask is executed,
even if nothing in the code changed. This happens because when you run
msbuild, it looks at the task's Inputs and Outputs, and if all inputs are
older than all outputs, then it skips the task. Since Inputs and Outputs
are not specified for BridgeNetCompilerTask, msbuild always performs this
task. I don't know how Visual Studio works.

As for the points you raised, you're right, I didn't look at the
configuration... Also, I only check for .js file, but don't check for .ts
file. Since I'm not familiar with the configuration, and it will take me a
while to do it properly (much longer than it would take to someone familiar
with your code), unfortunately, I can't spend the time to do a proper fix.
I hope someone on your end can do it.
For now, I'll just override the task's definition in my own repo, as a
workaround, in order not to wait ~40 seconds that BridgeNetCompilerTask
takes.

I do hope you fix it one day, in order to help others.
What's the best way for me to get notified when you release new versions of
the nuget?

geoffreymcgill:
> For now, I'll just override the task's definition in my own repo, as a workaround

This looks like a good work-around. 

Sorry we were not able to accept the PR. Unfortunately, not enough information was provided in order to reproduce the issue, and hard-coding of the `Inputs`/`Outputs` doesn't take into account user-specific settings based on the **bridge.json**. 

We will keep this performance issue in mind and hopefully another solution can be found.

